PR #8, opened 2023-02-01
[monitors] Implement BranchMonitor

btwj:
This monitor probes every `if`, `br_if`, and `br_table` instruction for the branch taken, and outputs the counts for each branch at the end of execution. It can be called with the `--monitors=branches` flag.


titzer (approved):
lgtm

PR #9, opened 2023-02-01
[monitors] Tree Profiling

btwj:
This PR implements a basic tree profiler, that shows a tree of function calls and the time spent within each function context (excluding function calls from within the function body).
- Called with `--monitors=profile` or `--monitors=profile{<depth>}`.
  - Example without depth:
    image
  - Example with depth `profile{2}`:
    image
- Caveats: can't use the normal testing setup for `profile_monitor0.wat`, because the execution times will vary.


titzer:
> This PR implements a basic tree profiler, that shows a tree of function calls and the time spent within each function context (excluding function calls from within the function body).
> 
> * Called with `--monitors=profile` or `--monitors=profile{<depth>}`.
>   
>   * Example without depth:
>     image
>   * Example with depth `profile{2}`:
>     image

Neat! Any ideas on how to display both the combined (nested + self) in an intuitive way? Or, at least, in a way more similar to existing tools?

> * Caveats: can't use the normal testing setup for `profile_monitor0.wat`, because the execution times will vary.

For that we could have a fuzzy match on the test output that just ignores the actual timings but checks the rest, e.g. maybe a regex.

btwj:
Hmm, from what I understand from looking at common profilers they seem to usually report only the combined (nested + self) time or they place the two numbers side by side (nested + self, self).

Side note but it may be cool to output some kind of flamegraph-like representation too, which could be built off the same infrastructure (but this won't be a sampling profiler like most flamegraphs).

titzer:
> Side note but it may be cool to output some kind of flamegraph-like representation too, which could be built off the same infrastructure (but this won't be a sampling profiler like most flamegraphs).

A flame graph sounds cool; maybe that can be an external tool and we can have an option to output the results in a machine-readable format. I think flame graphs are typically event-oriented and not summaries though?

btwj:
> I think flame graphs are typically event-oriented and not summaries though?
Yeah, I think they're usually representing chronological series of events. This monitor does a bunch of related work already though, such as tracking when functions are entered/exited and their call stack, which could be rewired to track times instead of durations. But I'll get this profiler working first before trying that out.

src/monitors/ProfileMonitor.v3
```diff
@@ -0,0 +1,196 @@
...
+    }
+    def enter() {
+        curTime = System.ticksUs();
+        count++;
+    }
+    def exit() {
+        totalTime += System.ticksUs() - curTime;
+    }
+
+    def render(T: TraceBuilder, m: Module, depth: int, total: int, maxDepth: int) {
```
titzer, line 25: I usually use `out` or `buf` for TraceBuilder variables.

src/monitors/ProfileMonitor.v3
```diff
@@ -0,0 +1,196 @@
+// Copyright 2022 Ben L. Titzer. All rights reserved.
+// See LICENSE for details of Apache 2.0 license.
+
+// A tree-based profiler
+class ProfileTreeNode(parent: ProfileTreeNode, id: int) {
+    var children: HashMap<int, ProfileTreeNode> = HashMap<int, ProfileTreeNode>.new(int.!<int>, int.==);
```
titzer, line 6: We'll need to eventually keep track of functions in other modules, so an `int` id won't be sufficient. Also, the overhead of a `HashMap` might be a lot, and I suspect most nodes have only a few children. So maybe just a vector or list of `WasmFunction`?

btwj: Done

src/monitors/ProfileMonitor.v3
```diff
@@ -0,0 +1,196 @@
...
+    }
+    private def renderRight(T: TraceBuilder, len: int, f: TraceBuilder -> void) {
+        var str = TraceBuilder.new();
+        f(str);
+        for (i < len - str.length) {
+            T.putc(' ');
+        }
+        T.puts(str.extract());
+    }
+    private def renderUs(T: TraceBuilder, time: int) {
```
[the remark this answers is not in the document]
btwj: I refactored this into a `TraceUtil.v3` file

src/monitors/ProfileMonitor.v3
```diff
@@ -0,0 +1,196 @@
...
+    }
+}
+
+class ProfileMonitor extends Monitor {
+    var tree: ProfileTree;
+    var maxDepth: int = -1;
+
+    def configure(args: string) -> string {
+        if (args.length > 0) {
+            maxDepth = Ints.parseDecimal(args, 0).1;
```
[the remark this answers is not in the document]
btwj: Done

src/util/TraceUtil.v3
```diff
@@ -0,0 +1,35 @@
+// Copyright 2023 Wizard Authors. All rights reserved.
+// See LICENSE for details of Apache 2.0 license.
+
+component TraceUtil {
+    // Renders a right-aligned string in a column of length `len`
+    def renderRight(out: TraceBuilder, len: int, f: TraceBuilder -> void) {
```
titzer, line 6: Today I added a utility method to `StringBuilder` (which `TraceBuilder` will then inherit): `rjustify`, which allows you to retro-actively insert spaces `N` characters back. I think you can make use of it here, or in the caller(s).

src/util/TraceUtil.v3
```diff
@@ -0,0 +1,35 @@
...
+        out.putd(intVal);
+        out.putc('.');
+
+        var base = 10;
+        for (i < prec) {
+            out.putd((dividend * base / divisor) % 10);
+            base *= 10;
+        }
+    }
+    def renderPct(out: TraceBuilder, dividend: long, divisor: long, prec: int) {
```
titzer, line 26: I think I prefer spelling this name out: `renderPercent`.

I'm a little inconsistent on this myself, so it'd probably be good for me to develop some kind of guideline on naming.

src/util/TraceUtil.v3
```diff
@@ -0,0 +1,35 @@
...
+    def renderRight(out: TraceBuilder, len: int, f: TraceBuilder -> void) {
+        var str = TraceBuilder.new();
+        f(str);
+        for (i < len - str.length) {
+            out.putc(' ');
+        }
+        out.puts(str.extract());
+    }
+    // Renders the decimal `dividend/divisor` to `prec` decimal places
+    def renderFrac(out: TraceBuilder, dividend: long, divisor: long, prec: int) {
```
titzer, line 15: spelling out "Fraction".

src/monitors/ProfileMonitor.v3
```diff
@@ -0,0 +1,184 @@
...
+        curNode.enter();
+    }
+    def exitFunc() {
+        curNode.exit();
+        curNode = curNode.parent;
+    }
+}
+
+class ProfileMonitor extends Monitor {
+    var tree: ProfileTree;
```
titzer, line 115: The indentation is a little off in this file. I always use one tab per indentation level.

titzer (approved):
lgtm with only minor nits

PR #12, opened 2023-02-09
[monitors] Create util for function enter/exit probes

btwj:
- Introduces `Module.insertProbeAtFunc{Enter|Exit}`, that attaches probes to a function's enter and exit.
- Works for functions that begin with a Loop Opcode by keeping a linked list of frame accessors. We need a linked list here instead of a single field in the case that a function that starts with a loop calls itself recursively.
- Refactors `CallsMonitor` and `ProfileMonitor` to use the new utils.


src/monitors/Monitor.v3
```diff
@@ -40,3 +40,23 @@ class Monitor {
...
+		var it = BytecodeIterator.new();
+		var bi = it.reset(func);
+		bi.current();
+		module.insertProbeAt(func.func_index, bi.pc, probe);
+	}
+	def probeOnExitFunc(module: Module, func: FuncDecl, probe: Probe) {
+		var it = BytecodeIterator.new();
+		for (bi = it.reset(func); bi.more(); bi.next()) {
+			var op = bi.current();
+			match (op) {
```
titzer, line 54: One oversight with the existing code I noticed is that branches to block `0` (i.e. the function scope) are not instrumented. I think this is why the calls monitor has been getting misaligned.

src/monitors/Monitor.v3
```diff
@@ -40,3 +40,23 @@ class Monitor {
 	}
 }
 
+component MonitorUtil {
+	def probeOnEnterFunc(module: Module, func: FuncDecl, probe: FuncEnterProbe) {
+		var it = BytecodeIterator.new();
+		var bi = it.reset(func);
+		bi.current();
+		module.insertProbeAt(func.func_index, bi.pc, probe);
```
titzer, line 48: I think we only want to use the `FrameAccessor` list trick if the first bytecode is `LOOP`, since it is more expensive than a regular probe (and creates garbage).

src/monitors/Monitor.v3
```diff
@@ -40,3 +40,23 @@ class Monitor {
 	}
 }
 
+component MonitorUtil {
```
titzer, line 43: Can we hide this as an implementation detail by introducing methods on the `Module` (such as `insertProbeBeforeEntry()` and `insertProbeBeforeReturn()`? That way the interface is uniform through the module and MonitorUtil is an implementation detail.

test/wizeng/loop_start.wat
```diff
@@ -0,0 +1,28 @@
+(module
+  (func $main (export "main")
+    (call $start (i32.const 10000))
```
titzer, line 3: 10000 seems like a lot. A few hundred would suffice?

titzer (changes requested):

src/monitors/Monitor.v3
btwj: Does Wizard have a nice way of getting the block depth in `BytecodeIterator`, since the branch label is relative? Or I could write some kind of `BytecodeVisitor` that keeps track of the depth.

This is getting pretty complicated though, especially having to get the values from the opstack for `br_if` and `br_table`s. Not sure if there's a better design you might have in mind.

titzer: It doesn't yet have such a utility; that would be pretty rad. Maybe a new class `BlockIterator`?

btwj: Makes sense. Might be useful in `LoopMonitor` too, where I was manually pushing and popping from a control stack to detect nested loops.

btwj: I'm not sure if this is consistent across execution modes, but it appears that branches to block `0` are already automatically instrumented because we added a probe to the final `end` bytecode. I think it's double-counting exits if I instrument both. Do branches to block `0` hit the final `end` based on your sidetable implementation or am I misunderstanding?

titzer: Ah yes, you are right. The intention was that they would in fact hit the final `END` bytecode. Probably worth adding an explicit test for this behavior.

src/monitors/FuncMonitorUtil.v3
```diff
@@ -0,0 +1,48 @@
+private class FuncEnterNoLoopProbe(probe: Probe) extends Probe {
+	var frameAccessor: FrameAccessor;
+	def fire(func: WasmFunction, pc: int, frame: TargetFrame) -> Resumption {
+		var curAccessor = frame.getFrameAccessor();
+		if (curAccessor == frameAccessor) return Resumption.Continue;
+		frameAccessor = curAccessor;
+		return probe.fire(func, pc, frame);
+	}
+}
+private class FuncEnterLoopProbe(probe: Probe) extends Probe {
```
titzer, line 10: I think you can avoid a search of the stack of accessors if you have the exit probe pop. So you have a matched pair of enter/exit helper probes for the loop case, and none for the no-loop case.

btwj: Makes sense, I'll try out that approach instead

titzer (approved):
lgtm, very nice!

PR #13, opened 2023-02-13
Add simple CI

k-sareen:


k-sareen:
Very confused why it doesn't work. `ocamlbuild` should theoretically be optional dependency based on the `Makefile` for the WASM spec. There is some issue with `progress` not working as well. Very odd.

k-sareen:
Aha it runs now: https://github.com/k-sareen/wizard-engine/actions/runs/4164314730.

Should be good to review now @titzer.

ci/linux/build-test-x86_64.sh
```diff
@@ -0,0 +1,26 @@
...
+cd virgil
+
+# Set up latest version of virgil
+export PATH=$PATH:"$PWD/bin:$PWD/bin/dev:$PWD/test/config"
+./test/configure
+make
+
+# Set up wizard
+cd ../wizard-engine
+make -j
```
titzer, line 15: There is an `x86-64-linux` target in the Makefile that will build only for that platform. It's a small savings.

test/all.sh
```diff
@@ -47,9 +62,15 @@ fi
 function do_script() {
     script=$1
     $SCRIPT_LOC/${script}.sh
+    X=$?
+    update_exit_code_if_non_zero $X
     if [[ "$TEST_TARGET" = "x86-64-linux" && $DEFAULT_MODES = 1 ]]; then
         TEST_MODE=jit $SCRIPT_LOC/${script}.sh
+        X=$?
+        update_exit_code_if_non_zero $X
```
titzer, line 70: I think there is a way to do this with `&&` in bash.

ci/jvm/build-test-jvm.sh
```diff
@@ -0,0 +1,26 @@
...
+export PATH=$PATH:"$PWD/bin:$PWD/bin/dev:$PWD/test/config"
+./test/configure
+make
+
+# Set up wizard
+cd ../wizard-engine
+make -j
+
+# Install OCaml for specification tests
+sudo apt install -y ocaml ocamlbuild
```
titzer, line 18: The `wasm-spec` repo seems to be pretty sensitive to the ocaml version. Do we need to specify that too?

titzer (commented):
Thanks for doing this! Looks mostly good to me.

ci/linux/build-test-x86_64.sh
k-sareen: Done

test/all.sh
k-sareen: Note sure what you mean. The point of the `X=$?` stuff is to make the CI fail if any of the tests fail. `&&` is generally used for chaining?

ci/jvm/build-test-jvm.sh
k-sareen: The `Makefile` in the `wasm-spec` repo says anything > 4.08 is fine. Ubuntu 22.04 has OCaml 4.13.1, so it should be okay unless they bump the min version requirement for OCaml in `wasm-spec`.

titzer: In theory, yes. In practice I have hit several incompatibilities (but mostly with proposal repos, which are forks of the main spec). I guess it's OK for now.

test/all.sh
titzer: I just mean, shortening up the code a little by not needing a whole `if` cascade. I think this might work:

`<shell command> || SCRIPT_EXIT_CODE=$?`

k-sareen: Ah right. I wanted to preserve the first error code as opposed to the last one, which is why I did this. Up-to-you if you want to preserve the first error code or the last one.

ci/jvm/build-test-jvm.sh
k-sareen: I haven't added proposal tests as I wasn't sure how to do that. I tried getting `opam` or whatever to work to get the latest OCaml version, but it was gloriously failing on the CI (while working on my local machine), so I gave up and just got the package manager versions of OCaml and ocamlbuild.

test/all.sh
titzer: Why not just exit early on the first error?

k-sareen: Hm, welll you get more data to work with if something fails. But yeah, we could also just do exit at first failure.

titzer: I think it's better to fail at first error in CI; that way you can always scroll to the end of the log to see what failed. (Come to think of it, it might be worth doing the same for the Virgil CI).

k-sareen: Right that's fair. I'll change it to fail on first error

k-sareen: Done.

PR #15, opened 2023-02-23
WIP: [monitors] Monitor Refactor

btwj:
Attempting to separate out `ModuleMonitor` and create a bunch of functions so most users of the interface won't have to write probes directly, only specify callbacks for specific events. Changed `BranchMonitor` and `CallsMonitor` to get a feel for this modified interface, and also added the ability to register memory events via this interface.
- Right now, each monitor is creating its own `ModuleMonitor` and registering its callbacks with its own `mm`. Should I move the creation of the monitor into engine code and change `onParse(module: Module)` to `onParse(moduleMonitor: ModuleMonitor)`? If the engine creates a single ModuleMonitor and passes the same copy to every monitor, this opens the possibility of multiple monitors sharing the same probes (e.g. if multiple different monitors instrument calls/returns, they can rely on the same function entry and exit probes instead of attaching their own)
- Opted for a more callback-heavy approach instead of having users create probes. Benefit is that many monitors end up creating probes that point back to themselves anyway. Creating callbacks in the `Monitor` definition as private functions means the callbacks naturally have access to monitor state.


btwj:
Hey @titzer, wanted some feedback on whether this direction for the refactor makes sense (it would hide a lot of the probe manipulation for most use cases) or if I should try a different approach.

titzer:
I generally like the direction of moving to first-class functions instead of the `Probe` class. We can bikeshed the names of things a bit. 

For the memory instrumentation, does it support narrowing the memory range of interesting events?

btwj:
I've added a way to only fire the callback when it's within a specified range

titzer:
Closing this one, as most of it has been integrated into others.

PR #16, opened 2023-03-06
[monitors] Convert ProbeFrame to DynamicLoc and change probe signature

btwj:
Also includes placeholder implementation for `V3Frame.caller()` (before we change the signature of `FrameAccessor.caller()` to support host functions too).


titzer (approved):
lgtm.

PR #17, opened 2023-03-08
[monitors] Create ModuleMonitor for Callback Utils

btwj:
Creates `ModuleMonitor` for certain callbacks that are built on top of probes.
- Includes `before/afterInstr`
- Refactoring existing monitors will be in a separate PR too.


src/engine/BytecodeIterator.v3
```diff
@@ -12,7 +12,7 @@
 class BytecodeIterator {
 	def var func: FuncDecl;
 	def var pc: int;
-	private def codeptr = CodePtr.new(null);
+	def codeptr = CodePtr.new(null);
```
titzer, line 15: I think we should add a `nextPC()` method to avoid directly exposing the code pointer here.

src/monitors/AfterInstrMonitor.v3
```diff
@@ -0,0 +1,31 @@
...
+
+class AfterInstrMonitor extends Monitor {
+	def configure(args: string) -> string {
+		return null;
+	}
+	def onParse(module: Module) {
+		var mm = ModuleMonitor.new(module);
+		mm.forEachFunc(MonitorUtil.isNotImport, probeEveryInstr(mm, _));
+	}
+	def onExit(code: int) {
```
titzer, line 12: Can probably remove because of the default implementation.

src/monitors/AfterInstrMonitor.v3
```diff
@@ -0,0 +1,31 @@
+// Copyright 2023 Wizard Authors. All rights reserved.
+// See LICENSE for details of Apache 2.0 license.
+
+class AfterInstrMonitor extends Monitor {
```
titzer, line 4: I think this name should be `PostInstructionTestTracer` or something similar. (and there should be a testcase for it), since it seems to mostly serve as a test case for the probe-after functionality.

src/monitors/ModuleMonitor.v3
```diff
@@ -0,0 +1,273 @@
+// Copyright 2023 Wizard Authors. All rights reserved.
+// See LICENSE for details of Apache 2.0 license.
+
+class ModuleMonitor(module: Module) {
```
titzer, line 4: We should have a couple of lines of top-level documentation for this class. Ultimately I think this will be the main way monitors are implemented, so it's good to have an explanation of that at the class level.

src/monitors/ModuleMonitor.v3
```diff
@@ -0,0 +1,273 @@
...
+	// Attach a callback that is called when func (non-import) is called
+	def beforeFuncCall(func: FuncDecl, f: DynamicLoc -> Resumption) {
```
titzer, line 6: Since these probes fire after the frame and arguments are set up (because it's done with an instruction probe in the target function), I think we should refer to this event as a "function execution" whereas a "function call" event happens in the caller, before the callee function's frame is set up. In that case, it'd be "beforeExecFunc"?

wdyt?

src/monitors/ModuleMonitor.v3
```diff
@@ -0,0 +1,273 @@
...
+	def beforeReturn(f: DynamicLoc -> Resumption) {
+		forEachFunc(MonitorUtil.isNotImport, beforeFuncReturn(_, f));
+	}
+	// Attach a callback to any non-import function call or return
+	def beforeCallAndReturn(callFn: DynamicLoc -> Resumption, retFn: DynamicLoc -> Resumption) {
+		forEachFunc(MonitorUtil.isNotImport, beforeFuncCallAndReturn(_, callFn, retFn));
+	}
+
+	// Attach a callback to an instruction
+	def beforeInstr(func: FuncDecl, pc: int, f: DynamicLoc -> Resumption) {
```
titzer, line 59: Similarly, I think we could name this "beforeExecInstr" for symmetry and clarity.

src/monitors/ModuleMonitor.v3
btwj: Yeah, that makes sense to me.

src/monitors/ModuleMonitor.v3
btwj: I ended up renaming them to `beforeFuncExec` and `beforeInstrExec` for symmetry with `beforeFuncReturn`, let me know if you think `beforeExecFunc` is better

btwj:
I'll remove the `FuncMonitorUtil` stuff too when I refactor the other monitors in a separate PR

PR #18, opened 2023-03-17
Implement more simd bitwise operations

haoyu-zc:
Including v128.not,  v128.andnot, v128.bitselect. All passed the spec tests.


titzer:
Thanks, can you please take a look at why the 'selfhost' tests failed--I believe these just need to be added to the unittest.failures file? Or it might be that the selfhost tests don't run with the failures file.

titzer (approved):
lgtm

PR #21, opened 2023-03-28
[monitors] Memory monitor

btwj:
Adds the most basic memory access callbacks to `ModuleMonitor`.
Doesn't support probes added directly to memories/pages yet.


titzer (approved):
lgtm

PR #22, opened 2023-04-04
[wizeng] Add wizeng module for test functions

btwj:
Adds a `wizeng` module for testing purposes.
- Added a `wizeng.invoke0` function that invokes a `funcref` from the host.

I had to add a function `invoke` to `Execute` that calls a function in the current execution context. `Execute.call` seems to reset everything and is used for entry, instead of calling a function in the middle of execution. I've only implemented it for `V3Interpreter` (with some duplicated code which I will refactor out), but wanted your feedback on if this was the right approach to things.


titzer (commented):
I think this is a fairly reasonable approach for now. I had wanted to make `Execute.call` clean up properly, but it was a bit difficult in the presence of exceptions, and IIRC I wanted to leave the stack intact for debugging. But I think it's best to clean up the stack by default, as it's less surprising. So doing that in the fast x86-64 tiers is just a matter of recording the value stack height and reseting it. Wdyt?

PR #27, opened 2023-04-20
[v3] Change representation of call stack in V3Interpreter

btwj:
Call stack is now represented as a doubly-linked list, with host frames and wasm frames inline.
- Allows frame accessors to get caller information even when their caller is a host function, and host functions can now get their callers as well.


src/engine/v3/V3Interpreter.v3
```diff
@@ -1330,14 +1317,91 @@ component V3Interpreter {
 		var r = f(double.view(x));
 		val_stack.elems[val_stack.top-1] = Value.F64(u64.view(r));
 	}
+
+	// Get the next frame, or make a new one if necessary
+	private def getNextFrame() -> V3Frame {
+		if (frame == null) {
+			frame = V3Frame.new();
+			frame.prev = null;
```
titzer, line 1325: These field initializations should be redundant.

src/engine/v3/V3Interpreter.v3
```diff
@@ -1330,14 +1317,91 @@ component V3Interpreter {
...
+			host_frame = host_frame.prev_host;
+		} else if (frame.func != null) {
+			call_stack_size--;
+		}
+		frame.hf = frame.func = frame.accessor = null;
+		frame = frame.prev;
+		return frame;
+	}
+	private def popFrames(prev_top: V3Frame) -> Array<(WasmFunction, int)> {
+		var frames: List<(WasmFunction, int)> = null;
```
titzer, line 1364: This will be more efficient if you can use a `Vector`, particularly if you can size it properly with `grow()`.

titzer (commented):
This is mostly lgtm. Somewhat mysterious that this is causing out-of-memory issues, as it doesn't really seem like frames being a few fields larger would trigger that.

src/engine/v3/V3Interpreter.v3
btwj: I realise now that I store the index of the frame in the frame itself, I can just pre-size the Array directly

btwj:
I'm still trying to figure out why there are out-of-memory issues; I had to reduce the call depth pretty substantially (~3000 instead of 10000) for it to pass. Might be a leak somewhere where some pointers aren't getting updated correctly and leaving some garbage live?

The new frames do seem to be bigger however, they have 7 pointer fields where the old frames only had 2 for the `WasmFunction` and `FrameAccessor`. I could reduce the overhead here by storing 3 of the pointer fields (`prev_host`, `next_host` and `hf`) on the side only for host frames since they're irrelevant for wasm frames, do you think that's worthwhile?

titzer:
> they have 7 pointer fields where the old frames only had 2

Hmm, that's only 5 extra pointers x 10000, which is about 200-400KB. That doesn't seem like the culprit. From the stack trace, it always seems to fail trying to grow the locals array. Are there locals accumulating across calls or something?

btwj:
I figured out the issue, the value stack wasn't getting reset properly when thrown.

PR #28, opened 2023-04-25
[simd]: Implement v128 integer comparison instructions

haoyu-zc:
Including:

- Equality
- Non-equality
- Less than
- Less than or equal
- Greater than
- Greater than or equal

(Need to accept the new PR for Virgil first)


src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2197,6 +2202,121 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
...
+			(Opcode.I32X4_GT_S, asm.pcmpgtd_s_s, GREATER),
+			(Opcode.I64X2_GT_S, asm.pcmpgtq_s_s, GREATER),
+
+			(Opcode.I8X16_LT_S, asm.pcmpgtb_s_s, LESS),
+			(Opcode.I16X8_LT_S, asm.pcmpgtw_s_s, LESS),
+			(Opcode.I32X4_LT_S, asm.pcmpgtd_s_s, LESS),
+			(Opcode.I64X2_LT_S, asm.pcmpgtq_s_s, LESS)
+		]) {
+			bindHandler(t.0);
+			if (t.2== GREATER) {
```
titzer, line 2235: I think it makes sense to factor out a method to load `r_xmm0` and `r_xmm1` at this point. For the comparisons where the order is reversed, another method that loads them in the other order.

titzer:
Merged this, can do the cleanup in a followup.

haoyu-zc:
Got it! Will create a new PR soon.

On Apr 30, 2023, at 17:46, Ben L. Titzer ***@***.******@***.***>> wrote:



Merged this, can do the cleanup in a followup.

PR #30, opened 2023-04-27
[monitors] CoverageMonitor uses basic block granularity

btwj:


src/monitors/CoverageMonitor.v3
```diff
@@ -4,18 +4,34 @@
 // Implements a simple monitor that tracks code coverage.
 class CoverageMonitor extends Monitor {
 	def entries = Vector<CoverageMonitorProbe>.new();
+	def isStartOfBlock = Vector<bool>.new();
 	def executed = Vector<bool>.new();
```
titzer, line 9: Can we keep a switch (even just a boolean in this file, as opposed to a parseable argument) that switches between per-instruction and per-basic block? I think we'll want to measure the performance of each.

titzer (approved):
lgtm

PR #31, opened 2023-05-10
[simd]: Refactor v128 comparison instructions

haoyu-zc:
Didn't figure out how to write lambdas in Virgil. So I put the helpers above genSimdInstrs()

### DIFF
--- a/src/engine/x86-64/X86_64Interpreter.v3
+++ b/src/engine/x86-64/X86_64Interpreter.v3
@@ -2062,6 +2062,16 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
 		}
 		asm.bind(call_irt);
 	}
+
+	// helper method for simd comparisons to move values from stack to registers
+	def simd_cmp_mov(lhs: X86_64Xmmr, rhs: X86_64Xmmr){
+		asm.movdqu_s_m(lhs, vsph[-2].value);
+		asm.movdqu_s_m(rhs, vsph[-1].value);
+	}
+	// reversed version of simd_cmp_mov, useful when dealing with "simd_cmp_mov_rev than" comparisons
+	def simd_cmp_mov_rev(lhs: X86_64Xmmr, rhs: X86_64Xmmr){
+		simd_cmp_mov(rhs, lhs);
+	}
 	def genSimdInstrs() {
 		bindHandler(Opcode.V128_LOAD); {
 			computeCurIpForTrap(-1);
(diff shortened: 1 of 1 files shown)


haoyu-zc:
Sure that makes sense! I'll make the change. And I have two more questions: 1. Does Virgil support Lambda? 2. Is it proper to put helper methods related to simd above the method genSimdInstructions or should I place them somewhere else?

On May 11, 2023, at 11:54, Ben L. Titzer ***@***.***> wrote:

﻿

@titzer commented on this pull request.

________________________________

In src/engine/x86-64/X86_64Interpreter.v3<https://github.com/titzer/wizard-engine/pull/31#discussion_r1191390627>:

> @@ -2062,6 +2062,16 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
                }
                asm.bind(call_irt);
        }
+
+       // helper method for simd comparisons to move values from stack to registers
+       def simd_cmp_mov(lhs: X86_64Xmmr, rhs: X86_64Xmmr){


I was thinking that if we name these something like "load_v128_xmm0_xmm1" and "load_v128_xmm1_xmm0", then we could use them in many more places.

titzer:
> 1. Does Virgil support Lambda?

It doesn't support lambdas, but it does support first-class functions (as you have seen) and partial applications of first-class functions (i.e. `foo(expr, _, _)`, which fixes some arguments and leaves others.

> 2. Is it proper to put helper methods related to simd above the method genSimdInstructions or should I place them somewhere else?

Yes, just above or below the SIMD instructions is fine.

src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2063,14 +2063,16 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
 		asm.bind(call_irt);
 	}
 
-	// helper method for simd comparisons to move values from stack to registers
-	def simd_cmp_mov(lhs: X86_64Xmmr, rhs: X86_64Xmmr){
-		asm.movdqu_s_m(lhs, vsph[-2].value);
-		asm.movdqu_s_m(rhs, vsph[-1].value);
+	// helper method to move values from vsp to registers
+	// vsp[-2] -> xmm0, vsp[-1] -> xmm1
+	def load_v128_xmm0_xmm1(xmm0: X86_64Xmmr, xmm1: X86_64Xmmr){
```
titzer, line 2068: Since this method is now always called with `(xmm0, xmm1)`, you should be able to remove the arguments.

haoyu-zc:
Removed the arguments.

titzer (approved):
lgtm

PR #35, opened 2023-05-16
[simd]: Implement v128 bit shifts

haoyu-zc:


haoyu-zc:
Could you rerun the CI again? I think it failed because my updates for Virgil had not merged.

PR #36, opened 2023-05-16
[monitors] Implement debugger using ModuleMonitor interface

btwj:
Implements a simple debugger interface using the `ModuleMonitor` interface.

Doesn't work outside interpreter mode for now as:
- Uses probes on interpreter loop
- `setLocal` not yet implemented for `X86_64BaseFrameAccessor`


titzer (approved):
lgtm, mostly rubber-stamped

PR #38, opened 2023-05-20
objdump new functions added

Katrina0406:
* Added new for loops to print out heaptypes, exports and so on besides functions in wasm programs.
* Added TraceOptions and MetricOptions to detect different flags in commands.

### DIFF
--- a/src/objdump.main.v3
+++ b/src/objdump.main.v3
@@ -11,7 +11,10 @@ def main(args: Array<string>) -> int {
 	for (i < args.length) {
 		var a = args[i];
 		if (a.length > 0 && a[0] == '-') {
-			if (!EngineOptions.parse(engine, a)) return ErrorBuilder.new()
+			if (!(TraceOptions.parse(a) 
+				|| EngineOptions.parse(engine, a)
+				|| MetricOptions.parse(a)
+				)) return ErrorBuilder.new()
 				.puts("objdump: unrecognized option ")
 				.putsq(a)
 				.ln()
@@ -59,5 +62,44 @@ def parseAndDump(engine: Engine, path: string) -> int {
 		if (f.imp == null) OUT.put1(" { %d bytes }", f.orig_bytecode.length);
 		OUT.outln();
 	}
+	for (i < module.heaptypes.length) {
+		var h = module.heaptypes[i];
+		OUT.put1("heaptypes[%d] = ", i);
+		h.render(OUT);
+		var fh = h.getFirstSuperType();
+		if (fh != null) fh.render(OUT);
+		OUT.outln();
+	}
+	for (i < module.abstypes.length) {
+		var a = module.abstypes[i];
+		OUT.put1("abstypes[%d] = ", i);
+		OUT.put1(": %z", a.isAssignableTo(ValueType.I32));
+		OUT.outln();
+	}
+	for (i < module.tables.length) {
+		var t = module.tables[i];
+		OUT.put1("tables[%d] = ", i);
+		OUT.put1(": has_default_elem %z", t.has_default_elem);
+		OUT.outln();
+	}
+	for (i < module.tags.length) {
+		var t = module.tags[i];
+		OUT.put1("tags[%d] = ", i);
+		if (t.fields != null) {
+			for (value in t.fields) {
+				OUT.put1("value: %q", value.render);
+				OUT.outln();
+			}
+		}
+		OUT.outln();
+	}
+	for (i < module.exports.length) {
+		var e = module.exports[i];
+		OUT.put1("exports[%d] = ", i);
+		OUT.puts(e.0);
+		OUT.outln();
+	}
+	OUT.put1("new functions: %d", module.new_funcs);
+	OUT.outln();
 	return 0;
 }


src/objdump.main.v3
```diff
@@ -59,5 +62,44 @@ def parseAndDump(engine: Engine, path: string) -> int {
...
+		OUT.put1("heaptypes[%d] = ", i);
+		h.render(OUT);
+		var fh = h.getFirstSuperType();
+		if (fh != null) fh.render(OUT);
+		OUT.outln();
+	}
+	for (i < module.abstypes.length) {
+		var a = module.abstypes[i];
+		OUT.put1("abstypes[%d] = ", i);
+		OUT.put1(": %z", a.isAssignableTo(ValueType.I32));
```
titzer, line 76: How about printing the types in the `assignable` array of each abs type?

src/objdump.main.v3
```diff
@@ -59,5 +62,44 @@ def parseAndDump(engine: Engine, path: string) -> int {
...
+		OUT.outln();
+	}
+	for (i < module.tables.length) {
+		var t = module.tables[i];
+		OUT.put1("tables[%d] = ", i);
+		OUT.put1(": has_default_elem %z", t.has_default_elem);
```
titzer, line 82: How about printing out the table's element type here?

src/objdump.main.v3
```diff
@@ -59,5 +62,44 @@ def parseAndDump(engine: Engine, path: string) -> int {
...
+			for (value in t.fields) {
+				OUT.put1("value: %q", value.render);
+				OUT.outln();
+			}
+		}
+		OUT.outln();
+	}
+	for (i < module.exports.length) {
+		var e = module.exports[i];
+		OUT.put1("exports[%d] = ", i);
```
titzer, line 98: How about printing out what kind of declaration and that declaration's index here?

titzer (changes requested):

src/objdump.main.v3
Katrina0406: I'm a bit confused about how to get the declaration type and index from Decl. I only saw the following two options
```
var imp: ImportInfo;
def imported() -> bool { return imp != null; }
```
And the ImportInfo class is empty. I tried using ".index" but it didn't work.

titzer: Oh right. The exports array has a string and a `Decl`. You can check the type of the decl to see what it is, e.g.

```
match (e.1) {
  x: FuncDecl => ...
  x: MemoryDecl => ...
  x: TableDecl => ...
  x: GlobalDecl => ...
}
```

src/objdump.main.v3
```diff
@@ -59,5 +62,88 @@ def parseAndDump(engine: Engine, path: string) -> int {
...
+		h.render(OUT);
+		var fh = h.getFirstSuperType();
+		if (fh != null) fh.render(OUT);
+		OUT.outln();
+	}
+	for (i < module.abstypes.length) {
+		var a = module.abstypes[i];
+		OUT.put1("abstypes[%d] = ", i);
+		if (a.isAssignableTo(ValueType.BOTTOM)) {
+			OUT.put1("%q ", ValueType.BOTTOM.render);
```
titzer, line 77: You should be able to just use the render method of `a`.

src/objdump.main.v3
```diff
@@ -59,5 +62,88 @@ def parseAndDump(engine: Engine, path: string) -> int {
...
+		}
+		if (a.isAssignableTo(ValueType.V128)) {
+			OUT.put1("%q ", ValueType.V128.render);
+		}
+		OUT.outln();
+	}
+	for (i < module.tables.length) {
+		var t = module.tables[i];
+		OUT.put1("tables[%d] default elem = ", i);
+		match (t.default_elem) {
```
titzer, line 99: Here, you should be able to use `t.name`, which is a string.

src/objdump.main.v3
```diff
@@ -59,5 +62,88 @@ def parseAndDump(engine: Engine, path: string) -> int {
...
+			}
+		}
+		OUT.outln();
+	}
+	for (i < module.exports.length) {
+		var e = module.exports[i];
+		OUT.put1("exports[%d] = ", i);
+		OUT.puts(e.0);
+		match (e.1) {
+			x: SigDecl => OUT.puts(": SigDecl");
```
titzer, line 133: Here, only functions, tables, memories, globals, and tags should be exportable. We can use the short names, "func", "table", "memory", "global", "tag", plus their respective index. (The reason for that is that `FuncDecl` and so on are names internal to Wizard, while the objdump utility should speak in terms that are WebAssembly-universal.).

titzer (approved):
Nice improvement. Last round of comments, and we should be good to go.

src/objdump.main.v3
Katrina0406: I'm a bit confused. Where can we find the render method of a? I checked AbsTypeDecl class and Decl class but didn't find anything related.

src/objdump.main.v3
Katrina0406: It gives me error:
```
[src/objdump.main.v3 @ 99:28] UnresolvedMember: expression of type TableDecl has no such member "name"
		OUT.puts(t.name);
```
Also, do you mean I should replace the entire match block with one line of t.name?

src/objdump.main.v3
Katrina0406: I'm confused by this line:

> We can use the short names, "func", "table", "memory", "global", "tag", plus their respective index.
Where do we use the short names? Is it inside the puts or we can just write Func instead of FuncDecl after the colon to represent type? Also, I didn't see where we can access their index. I saw that there's an import_index param in input of class ImportInfo but nothing inside the class can be used..?

titzer: Just inside the string, so that the output doesn't say "FuncDecl", etc.

titzer (approved):
I'll land this and make a few small adjustments.

PR #43, opened 2023-05-30
[simd]: Implement i32x4_extract_lane

haoyu-zc:


src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2094,6 +2094,24 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
 	def load_v128_xmm1_xmm0(){
 		load_v128_s_s(r_xmm1, r_xmm0);
 	}
+	def extract_v128_lane(opcode: Opcode, dst: X86_64Addr, src: X86_64Gpr, idx: X86_64Gpr) {
```
titzer, line 2097: As discussed, I think this helper should be something like `genExtractLane` and take the opcode and the size, and do the `bindHandler`, read the immediate byte, etc. Then the call sites can pass the specific opcode and size, no need to match on the opcode.

PR #47, opened 2023-06-06
[simd](bug): TypeCheckException when using null

haoyu-zc:


titzer:
I can reproduce this now and have a simpler failing test that I've added to the Virgil repo.

PR #48, opened 2023-06-06
[simd]: Fix local.set and local.tee for v128 values

haoyu-zc:


titzer:
I merged this, checked locally that it passes, but it fails in CI. I'll have to debug it.

haoyu-zc:
Thanks! I’ll work on other stuff for the time being

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

PR #49, opened 2023-06-08
Print diff as part of CI diagnosis

titzer:
Creating PR to create a workflow run to see diff


titzer:
For the record, this run uncovered that the CI was failing because the pregen code space was not large enough, and the `mprotect()` for read-execute made part of the initialized heap immutable, which caused a crash.

PR #50, opened 2023-06-09
[simd]: Implement v128 load_lane instructions

haoyu-zc:
- Refactored current `load_memarg` helpers to take in an `src` address to specify the memory location to be loaded
- Factored out `checkLaneAndSig` since `checkLaneLoad` and `checkLaneStore` already check the signature.


titzer (approved):
lgtm

PR #53, opened 2023-06-12
[simd]: Implement v128 extended pairwise integer addition

haoyu-zc:


src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2106,6 +2106,12 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
 	def load_v128_xmm1_xmm0(){
 		load_v128_s_s(r_xmm1, r_xmm0);
 	}
+	def load_v128_s(s: X86_64Xmmr) {
```
titzer, line 2109: I think it's clearer to just inline these for now and we can do a cleanup pass to factor out common patterns when all instructions are done.

haoyu-zc: Got it! Pushed a new commit

titzer (approved):

PR #54, opened 2023-06-12
[simd]: Implement v128 store and store_lane

haoyu-zc:
- Added an enum for `genStoreLane()`. I think this makes the caller side code more clear and concise since no assembler methods need to be passed around any more.
- If this code review passes, I'll refactor previous gen methods like `genStoreLane()` to follow this pattern


src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2202,8 +2206,46 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
 		if (valuerep.tagged) genTagUpdate(BpTypeCode.V128.code);
 		endHandler();
 	}
+	def genStoreLane(opcode: Opcode, size: MemargSize) {
```
titzer, line 2209: I think it's shorter to keep using the pattern of passing in the load/store methods.

haoyu-zc:
Updated!

PR #56, opened 2023-06-14
[simd]: Implement v128 splat and splat_load instructions

haoyu-zc:
I used  a `match` on the opcode when implementing these two sets of instructions, because:
- splats that with different value sizes have different implementations in `wasm`, i.e., the wasm functions take in different types of parameters so I'm not able to generalize them like what I did in `load_lane` and `store_lane`
- `load_and_splat` cases fall through to `splat` cases naturally. They share the same logic except for where the value is loaded from.


src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2220,6 +2220,43 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
 		asm_mov_m_r(mem_addr, val);
 		endHandler();
 	}
+	def genV128Splat(opcode: Opcode, from_mem: bool,
```
titzer, line 2223: I think this will be cleaner if you either split it to two functions (one from memory and one not). Further, it will be simpler to also just pass the `MacroAssembler` method for the actual splat, rather than switching again on the opcode.

haoyu-zc: I refactored `genSplat` to take in a `gen` method as an additional argument to avoid switching on the opcode.

src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2220,6 +2220,38 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
...
+	}
+	def gen_f32X4_splat() {
+		asm.pinsrd_s_r_i(r_xmm1, r_tmp0, 0);
+		masm.emit_f32x4_splat(r_xmm0, r_xmm1);
+	}
+	def gen_f64X2_splat() {
+		asm.pinsrq_s_r_i(r_xmm1, r_tmp0, 0);
+		masm.emit_f64x2_splat(r_xmm0, r_xmm1);
+	}
+	def genSplat(opcode: Opcode, from_memarg: bool,
```
titzer, line 2243: If you further split this method into two versions (`from_memarg=true` and `from_memarg=false`), then you can just pass methods from the `MacroAssembler` to it.

titzer: You can just bind a temporary register for the one that takes a scratch register, e.g. `masm.emit_i8x16_splat(_, _, r_tmp2)`.

haoyu-zc: Yeah I just forgot to use the partial application feature of Virgil. It's so convenient. Pushed a new change.

titzer (approved):
lgtm, thanks

PR #58, opened 2023-06-20
[simd]: Implement v128 swizzle

haoyu-zc:
- Also refactored masm.load_v128_mask to use one fewer xmmr


src/engine/x86-64/X86_64MacroAssembler.v3
```diff
@@ -963,6 +961,11 @@ class X86_64MacroAssembler extends MacroAssembler {
 		asm.movq_s_r(scratch, src);
 		asm.movddup_s_s(dst, scratch);
 	}
+	def emit_i8x16_swizzle(dst: X86_64Xmmr, src: X86_64Xmmr, tmp: X86_64Gpr, mask: X86_64Xmmr) {
```
titzer, line 964: Does this implementation handle out-of-bounds cases properly? 

(https://emscripten.org/docs/porting/simd.html?highlight=exception)

haoyu-zc: For 2, actually the cases are handled but not covered by tests. Adding new tests

haoyu-zc:
Closing this PR. Creating a new one for swizzle alone

PR #62, opened 2023-07-01
[simd]: Implement v128 bit shifts

haoyu-zc:
- Wrapping up implementing x86-64 simd by testing wizard against the proposal tests. More PRs on the way to fix bugs and add a few missing instructions


titzer (approved):
lgtm

PR #67, opened 2023-07-02
Inlining counter probes

btwj:
- `ProbeList` now extends `Probe`, so `Probe`s and `ProbeList` can both be fired in the same way.

I have both potential optimizations turned off for now, since they cause some tests to fail:
- the `CountProbe` optimization works until there's a GC, since we don't update the pointer of the `count` field in the JIT'd code right now
- the operand probe optimization doesn't work for now, haven't had time to fix it but some register is getting overwritten


titzer:
I've landed a version of this PR that adjust GC-robustness for both `CountProbe` and `OperandProbe_i_v`, but didn't yet debug the later.

PR #80, opened 2023-07-19
Implement simple f32x4 arithmetic operations

haoyu-zc:
Tested by ` make -j && bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_f32x4_arith.bin.wast`


titzer (approved):
lgtm

PR #81, opened 2023-07-20
[simd/jit]: Implement simple f64x2 arithmetic operations

haoyu-zc:
Tested by `make -j && bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_f64x2_arith.bin.wast`


titzer (approved):
lgtm

PR #82, opened 2023-07-21
[simd/jit]: Implement i8x16 comparison instructions

haoyu-zc:
Tested by `make -j && bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_i8x16_cmp.bin.wast`


titzer (approved):
lgtm

PR #83, opened 2023-07-21
[simd/jit]: Implement v128.const

haoyu-zc:
Tested with `make -j && bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_const.bin.wast`


titzer (approved):
lgtm

PR #84, opened 2023-07-22
[simd/jit]: Implement i16x8 comparison instructions

haoyu-zc:
Tested by `make -j && bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_i16x8_cmp.bin.wast`


titzer (approved):
lgtm, though would prefer to add expected failures to a file rather than commenting them out.

haoyu-zc:
True that will be better! Will do so in the future.

PR #85, opened 2023-07-22
[simd/jit]: Implement i32x4 comparison instructions

haoyu-zc:
Tested with `make -j && bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_i32x4_cmp.bin.wast`


titzer (approved):
lgtm

PR #86, opened 2023-07-22
[simd/jit]: Implement i64x2 comparison instructions

haoyu-zc:
Tested by `make -j && bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_i64x2_cmp.bin.wast`


titzer (approved):
lgtm

PR #87, opened 2023-07-23
Factor out i64x2_mul

haoyu-zc:


titzer (approved):
lgtm, nice simplification.

PR #90, opened 2023-07-28
Update CI scripts for specification tests

haoyu-zc:


haoyu-zc:
Testing on my repo now.

PR #92, opened 2023-07-29
[simd/jit]: Implement more i8x16 arithmetic instructions

haoyu-zc:
Tested by `make -j && bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_i8x16_arith2.bin.wast`


src/engine/x86-64/X86_64SinglePassCompiler.v3
```diff
@@ -612,6 +619,13 @@ class X86_64SinglePassCompiler extends SinglePassCompiler {
 		state.push(sv.kindFlagsMatching(kind, IN_REG), d, 0);
 		return true;
 	}
+	// r1 = op(r1), SIMD unop
+	private def do_op1_x<T>(kind: ValueKind, emit: (X86_64Xmmr) -> T) -> bool {
+		var sv = popReg(), r = X(sv.reg);
```
titzer, line 624: You need a `popRegToOverwrite()` if the code modifies the register. (that invalidates the other uses of the register, whereas `popReg()` does not).

titzer: Yeah, unfortunately these kind of register allocation bugs require testing each instruction under multiple different scenarios (i.e. the input is a constant, is in a register, is not a in a register, is in a register that is shared with another variable, etc). There is at least a limited number of scenarios, so maybe we could add a testing mode for it.

titzer (approved):

PR #104, opened 2023-08-03
[simd/jit]: Factor out do_op2_x_x_xtmp() and do_c_op2_x_x_xtmp()

haoyu-zc:
Tested by:

```bash
function jit_regress {
  make -j

  # Array of variables
  var_list=("i8x16" "i16x8" "i32x4" "i64x2" "f32x4" "f64x2")

  # Loop over array and execute command
  for var in "${var_list[@]}"
  do
      bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_${var}_arith.bin.wast
      # Only integer types have the second test
      if [[ ${var:0:1} != "f" ]]
      then
          bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_${var}_arith2.bin.wast
      else
          bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_${var}.bin.wast
      fi
  done | progress
}

jit_regress
```


titzer (approved):
lgtm

PR #106, opened 2023-08-05
[simd/jit]: Implement floating point rounding instructions

haoyu-zc:
Tested by
- `make -j && bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_f32x4_rounding.bin.wast`
- `make -j && bin/spectest.x86-64-linux -tk -mode=spc test/regress/simd/simd_f64x2_rounding.bin.wast`


titzer (approved):
lgtm

PR #123, opened 2023-10-17
[simd/v3i]: Implement more integer arithmetic instructions

haoyu-zc:
Tested by:
```
make v3i -j && bin/spectest.v3i -tk test/regress/simd/simd_i64x2_arith2.bin.wast
make v3i -j && bin/spectest.v3i -tk test/regress/simd/simd_i32x4_arith2.bin.wast
make v3i -j && bin/spectest.v3i -tk test/regress/simd/simd_i16x8_arith2.bin.wast
make v3i -j && bin/spectest.v3i -tk test/regress/simd/simd_i8x16_arith2.bin.wast
```

### DIFF
--- a/src/engine/V3Eval.v3
+++ b/src/engine/V3Eval.v3
@@ -498,259 +498,260 @@ component V3Eval {
 			_ => return val;
 		}
 	}
-	// Private (i.e. file-scoped) utilities.
-	// ---- v128 arithmetic helpers ----------------------------------------
-	def F32_ADD_U = float_binop(_, _, float.+);
-	def F32_SUB_U = float_binop(_, _, float.-);
-	def F32_MUL_U = float_binop(_, _, float.*);
-	def F32_DIV_U = float_binop(_, _, float./);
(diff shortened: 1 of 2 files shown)


haoyu-zc:
Rebased accordingly. But I'm not sure which utilities should be modified by private since I noticed you removed some `private`s in your latest commit. For methods such as `do_vv_v_x2`, I think they should be private bc other files may not use them, right?

titzer:
Functions at the top level of a file (not in a class or a component) are private by default.

src/engine/V3Eval.v3
```diff
@@ -441,108 +498,259 @@ component V3Eval {
 			_ => return val;
 		}
 	}
-}
+	// Private (i.e. file-scoped) utilities.
```
titzer, line 501: Since these are outside the component, they don't need to be indented another level.

haoyu-zc: Oh, I got it. Didn't notice they were moved outside of the component. Pushed a new change.

titzer (approved):

PR #126, opened 2023-10-27
[v3i/simd]: Implement comparison instructions

haoyu-zc:


src/engine/V3Eval.v3
```diff
@@ -609,6 +785,36 @@ def I8_POPCNT(a: u8) -> u8 {
 	return count;
 }
 // Adapters
+def u8_bool_binop(a: u8, b: u8, f: (u8, u8) -> bool) -> u8 {  // Adapts a unsigned bool binop to a u8 binop
```
titzer, line 788: What about `do_bb_z` as a name here? We haven't used an abbreviation for bytes yet, but we could use `b` for byte, `w` for u16, and `q` for v128. I use `z` for boolean in, e.g. Virgil's `lib/util`.

src/engine/V3Eval.v3
```diff
@@ -317,6 +317,24 @@ component V3Eval {
 	def I64X2_ABS(a: (u64, u64)) -> (u64, u64) {
 		return do_v_v_x2(a, V128_I64_ABS);
 	}
+	def I64X2_EQ(a: (u64, u64), b: (u64, u64)) -> (u64, u64) {
```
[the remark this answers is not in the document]
haoyu-zc: Yes for sure. I'll create a new PR for this.

src/engine/V3Eval.v3
haoyu-zc: Yeah the naming of these helpers did bother me a while. Thanks for the suggestion and I'll push a new commit for this.

haoyu-zc: Just to double-check: for v128, should we use `dq` as what we did in the x86_64Interpreter? And, accordingly, use `d` for u32 and `q` for u64?

titzer: I would propose this:

```
b = byte = u8
h = u16
i = i32
u = u32
l = i64
w = u64
f = float
d = double
q = v128
```

haoyu-zc: Yes this looks clean. Thanks! Pushing a new commit soon.

haoyu-zc:
For helpers take narrower types such as `i8`, I used a suffix to name them, e.g.:

`def do_i_i_8(a: u8, f: i8 -> i8) -> u8 {  // Adapts a signed i8 unop to a u8 unop`

If you prefer any other forms, let me know and I'll change them accordingly.

titzer (approved):

PR #128, opened 2023-10-30
[simd/v3i]: Implement saturating arithmetic instructions

haoyu-zc:


src/engine/V3Eval.v3
```diff
@@ -644,6 +646,11 @@ def I16_SUB_SAT_U(a: u16, b: u16) -> u16 {
 	if (dif < 0) return u16.view(0);
 	else return u16.view(dif);
 }
+def I16_Q15_MUL_SAT_S(a: i16, b: i16) -> i16 {
+	var prod = (i32.view(a) * i32.view(b) + 0x4000) >> 15;
+	if (prod > i16.max) return i16.max;
```
titzer, line 651: You can turn this `if` into a one-liner with `return if(prod > i16.max, i16.max, i16.view(prod));`

titzer (commented):
lgtm with minor simplification

src/engine/V3Eval.v3
haoyu-zc: Yes that's a ternary statement here. Will push a change.

But I'm not sure why the CI failed. Will take a closer look tmw.

titzer: The CI failed because of a breaking change. Should be fixed now.

src/engine/V3Eval.v3
```diff
@@ -588,6 +602,54 @@ def I8_POPCNT(a: u8) -> u8 {
...
 }
+def I8_ADD_SAT_S(a: i8, b: i8) -> i8 {
+	var sum = i16.view(a) + i16.view(b);
+	if (sum > 127) return 127;
+	else if (sum < -128) return -128;
+	else return i8.view(sum);
+}
+def I8_ADD_SAT_U(a: u8, b: u8) -> u8 {
+	var sum = u16.view(a) + u16.view(b);
+	if (sum > 255) return 255;
```
titzer, line 613: Can you also do the `return if(` here and elsewhere in this file?

haoyu-zc: Yes just found out I should do that after the new commit. Pushing a new one soon.

haoyu-zc:
Found more places could be simplified with `if (cond, a, b)`. A new commit is pushed

titzer (approved):

PR #129, opened 2023-11-02
[v3i/simd]: Implement min, max, abs instructions for f32 and f64

haoyu-zc:
Tested by:
```
make v3i -j && bin/spectest.v3i -tk test/regress/simd/simd_f32x4.bin.wast
make v3i -j && bin/spectest.v3i -tk test/regress/simd/simd_f32x4_pmin_pmax.bin.wast
make v3i -j && bin/spectest.v3i -tk test/regress/simd/simd_f64x2.bin.wast
make v3i -j && bin/spectest.v3i -tk test/regress/simd/simd_f64x2_pmin_pmax.bin.wast
```

### DIFF
--- a/src/engine/V3Eval.v3
+++ b/src/engine/V3Eval.v3
@@ -385,6 +385,11 @@ component V3Eval {
 	def F32X4_LE = do_vv_v_x4(_, _, V128_F32X4_LE);
 	def F32X4_GT = commute_binop(F32X4_LT);
 	def F32X4_GE = commute_binop(F32X4_LE);
+	def F32X4_MIN = do_vv_v_x4(_, _, V128_F32X4_MIN);
+	def F32X4_MAX = do_vv_v_x4(_, _, V128_F32X4_MAX);
+	def F32X4_ABS = do_v_v_x4(_, V128_F32X4_ABS);
+	def F32X4_PMIN = do_vv_v_x4(_, _, V128_F32X4_PMIN);
+	def F32X4_PMAX = do_vv_v_x4(_, _, V128_F32X4_PMAX);
 	def F64X2_ADD = do_vv_v_x2(_, _, F64_ADD_U);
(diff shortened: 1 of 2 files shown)


src/engine/V3Eval.v3
```diff
@@ -588,6 +608,26 @@ def I8_POPCNT(a: u8) -> u8 {
...
+	if (a <= b) return a;
+	if (a > b) return b;
+	return a;
+}
+def F32X4_PMAX(a: float, b: float) -> float {
+	if (a >= b) return a;
+	if (a < b) return b;
+	return a;
+}
+def F64X2_PMIN(a: double, b: double) -> double {
```
titzer, line 621: Can this be simplified to `if(a > b, b, a)`?

haoyu-zc: Take this one for example:
```
def F32X4_PMAX(a: float, b: float) -> float {
	if (a >= b) return a; // line1
	if (a < b) return b; // line2
	return a; // line3
}
```
When either a or b is a NAN, the control flow will fall through to line 3. And we need to return the first operand here according to the test cases. (a and b could be a quiet NaN or a signaling NaN, having different bit patterns).

So I'm not able to simplify it to `return if(a >= b, a, b);` as the tests will fail.

haoyu-zc: Take this one for example:
```
def F32X4_PMAX(a: float, b: float) -> float {
	if (a >= b) return a; // line1
	if (a < b) return b; // line2
	return a; // line3
}
```
When either a or b is a NAN, the control flow will fall through to line 3. And we need to return the first operand here according to the test cases. (a and b could be a quiet NaN or a signaling NaN as I observed).

So I'm not able to simplify it to `return if(a >= b, a, b);` as the tests will

haoyu-zc:
Alright I turned my head around... Pushed a new commit.

titzer (approved):
lgtm

PR #134, opened 2023-11-10
[v3i/simd]: Implement bit shifts

haoyu-zc:


src/engine/V3Eval.v3
```diff
@@ -823,6 +861,19 @@ def do_v_v_x16(a: (u64, u64), f: (u8) -> u8) -> (u64, u64) { // Performs a 16-la
 	}
 	return (low, high);
 }
+// f(v128, int) -> v128
+def do_vi_v_x16(a: (u64, u64), b: i32, f: (u8, u8) -> u8) -> (u64, u64) {
```
titzer, line 865: These are going to do a partial application for each invocation. Can you find a way to refactor so that the partial applications only happen when initializing the publicly-exposed methods?

haoyu-zc: Yes I realized there will be extra runtime overheads if I do things this way. But I haven't figured out an elegant way to refactor the code. For now I change the helpers such as `do_vi_v_x16` to handle the logic themselves to avoid partial applications. Will keep on thinking about how to improve while I'm working on other instructions.

src/engine/V3Eval.v3
```diff
@@ -823,6 +861,47 @@ def do_v_v_x16(a: (u64, u64), f: (u8) -> u8) -> (u64, u64) { // Performs a 16-la
...
+		r_a = u8.view(a.1 >> shift);
+		res = f(r_a, u8.view(b));
+		high |= (u64.view(res) << shift);
+	}
+	return (low, high);
+}
+def do_vi_v_x8(a: (u64, u64), b: i32, f: (u16, u16) -> u16) -> (u64, u64) {
+	var low: u64 = 0;
+	var high: u64 = 0;
+	for (shift: byte = 0; shift < 64; shift += 16) {
```
titzer, line 882: This will be a little more efficient if you just loop with a regular `int` induction variable and then do a `u6.view` once for each iteration.

haoyu-zc: Updated. Also fix places where code is sub-optimal.

PR #136, opened 2023-11-13
[v3i/simd]: Implement v128 extract_lane instructions

haoyu-zc:
Tested by `make v3i -j && bin/spectest.v3i -tk test/regress/simd/v128_extract_lane.bin.wast`

### DIFF
--- a/src/engine/v3/V3Interpreter.v3
+++ b/src/engine/v3/V3Interpreter.v3
@@ -914,10 +914,10 @@ component V3Interpreter {
 				if (t != TrapReason.NONE) trap(t);
 			}
 			V128_LOAD => doLoad(16, DataReaders.read_range_u128, Value.V128);
-			V128_LOAD_64_LANE => doLoadLane(8, DataReaders.read_range_u64);
-			V128_LOAD_32_LANE => doLoadLane(4, DataReaders.read_range_u32);
-			V128_LOAD_16_LANE => doLoadLane(2, DataReaders.read_range_u32_u16);
-			V128_LOAD_8_LANE => doLoadLane(1, DataReaders.read_range_u32_u8);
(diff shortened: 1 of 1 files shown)


src/engine/v3/V3Interpreter.v3
```diff
@@ -1105,6 +1113,21 @@ component V3Interpreter {
 		}
 		push(Value.V128(low, high));
 	}
+	def doExtractLane<T, S>(size: byte, view: u64 -> T, extend: T -> S, stack_push: (S) -> void) {
```
titzer, line 1116: Can you pass in the log-base-2 of the size instead of the size, so that the divide and multiplies become shifts?

src/engine/v3/V3Interpreter.v3
```diff
@@ -1105,6 +1113,21 @@ component V3Interpreter {
...
+		var v = Values.v_v128(pop());
+		var idx = codeptr.read1();
+		var low = v.0, high = v.1;
+		var val: S;
+		def half_lanes = (16 / size) >> 1;
+		if (idx < half_lanes) { // Extract a lane from low
+			var shift = byte.view(idx * size * 8);
```
titzer, line 1123: Here (and elsewhere), it's a tiny bit more efficient to do a `u6.view(x)` to compute a shiftor (RHS of shift), because Virgil semantics are that the shiftor is bounds-checked against the width of the type, and `u6` is always in bounds for `u64`.

haoyu-zc: Ah that makes perfect sense since we only have 2^6 bits to shift within a `u64`. Will update accordingly.

src/engine/v3/V3Interpreter.v3
haoyu-zc: Sure!

haoyu-zc:
Updates:
- Use log-base-2 sizes for strength reduction
- Use narrower shiftors to improve perf

PR #137, opened 2023-11-16
[v3i/simd]: Implement v128 splat and load_splat

haoyu-zc:
Tested by:
```
make v3i -j && bin/spectest.v3i test/regress/simd/simd_splat.bin.wast
make v3i -j && bin/spectest.v3i test/regress/simd/simd_load_splat.bin.wast
```


titzer (approved):
lgtm

PR #138, opened 2023-11-16
[ci] Update to include menhir and its dependencies

haoyu-zc:
We need to add `-y` since menhir depends on other libraries, and opam asks for our permission to proceed.

Successful CI run in my repo: https://github.com/haoyu-zc/wizard-engine/actions/runs/6894577687/job/18756696731


titzer (approved):
lgtm, thanks!

PR #148, opened 2023-11-27
[simd/v3i]: Implement v128 shuffle and swizzle

haoyu-zc:
The final piece of the SIMD puzzle is solved.

### DIFF
--- a/src/engine/v3/V3Interpreter.v3
+++ b/src/engine/v3/V3Interpreter.v3
@@ -1274,10 +1274,10 @@ component V3Interpreter {
 	}
 	def doShuffle0(a: (u64, u64), b: (u64, u64)) -> u64 {
 		var res = 0ul;
-		var f = doExtractLane0(_, _, 0, u8.view<u64>, u8.view<u8>);
 		for (i < 8) {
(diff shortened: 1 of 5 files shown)


src/engine/v3/V3Interpreter.v3
```diff
@@ -1258,7 +1263,42 @@ component V3Interpreter {
...
+	def doShuffle(){
+		var b = Values.v_v128(pop());
+		var a = Values.v_v128(pop());
+		var low = doShuffle0(a, b);
+		var high = doShuffle0(a, b);
+		push(Value.V128(low, high));
+	}
+	def doShuffle0(a: (u64, u64), b: (u64, u64)) -> u64 {
+		var res = 0ul;
+		var f = doExtractLane0(_, _, 0, u8.view<u64>, u8.view<u8>);
```
titzer, line 1277: Can we remove this partial application here? It will allocate a closure every time the bytecode is executed.

haoyu-zc: Yes I found this problem yesterday and pushed a change. But I did a force push later by mistake so it was overwritten... Updated.

PR #149, opened 2023-12-31
[fast-int]: Clean up code for v128 instructions

haoyu-zc:
- Factored out more tuple loops
- Created more generic routines
- Moved instructions around for better grouping


titzer (approved):
Nice! lgtm

PR #153, opened 2024-02-19
Add memory module for shared memories.

aoli-al:
This PR introduces the shared memory module to Wizard. 

- We need to make `ModuleRegistry` more generic to accept `MemoryModule`
- Implement `MemoryModule`
- Add wasm tests


src/modules/ModuleRegistry.v3
```diff
@@ -23,6 +23,23 @@ component ModuleRegistry {
 		if (t.0 != null) map[name] = (t.0, true);
 	}
 }
+
+// A memory module used for creating shared memories.
```
titzer, line 27: This seems like a good candidate for factoring out to it's own file.

src/modules/wasi/WspOneModule.v3
```diff
@@ -329,6 +329,7 @@ def OOB = HostResult.Throw(Trap.new(TrapReason.MEM_OUT_OF_BOUNDS, null, null));
 // Enforces ordering on installation of platform-specific modules.
 component WspOneRegistration {
 	def X = ModuleRegistry.add(WspOneModule.new(), true); // register platform-independent first
+	def M = ModuleRegistry.add(MemoryModule.new("env"), true); // Only register `env::memory` when WASI is enabled.
```
titzer, line 332: Since this "env.memory" is actually part of WASI, what about refactoring so that the `WspOneModule` supplies this import? (I.e. no separate `MemoryModule`)?

titzer: Yeah, it seems like WASI is a special kind of thing that shouldn't just be keyed by the name of the imported module. I think it makes sense to make it an `ImportProcessor` that runs over imports, controlled by a separate flag.

aoli-al: Ah I see. This sounds good to me. Will work on this.

PR #156, opened 2024-04-09
Add path resolver for wasi

yicheng4:


src/modules/wasi/x86-64-linux/WspOnePathResolver-x86-64-linux.v3
```diff
@@ -2,9 +2,29 @@
 // See LICENSE for details of Apache 2.0 license.
 component WasiPathUtil {
 	def normalize_path = normalize_path0;
+	def resolve_path = resolve_path0;
 
 }
 
+type PathComponent {
```
titzer, line 9: I think you can remove this.

src/modules/wasi/x86-64-linux/WspOnePathResolver-x86-64-linux.v3
```diff
@@ -2,9 +2,29 @@
...
+def SYS_readlink = 0x59;
+
+// Maximum allowed path length.
+def PATH_MAX = 4096;
+
+def EINVAL = 22;
+def ENOENT = 2;
+
+// Checks if a given path is absolute by looking for a leading slash.
 def is_absolute_path(path: Array<byte>) -> bool {
```
titzer, line 28: In Virgil, `string` and `Array<byte>` are equivalent types (right now), but the intention is that strings are immutable in the future. I think it makes sense to use `string` where an immutable string is expected; it's a bit shorter and also future-compatible.

titzer (approved):
lgtm

PR #161, opened 2024-05-11
[WIP] [stack-switching] Bytecode validation for 

davidmaamoaix:
This is a work-in-progress PR for implementing code validation for the new instructions introduced in the stack-switching proposal.

This PR also updates the bytecode parsing for stack-switching. `ContDecl` now contains a direct reference to its corresponding `SigDecl` instead of an index to it in the `Module.heaptypes` array. This allows context-free check for subtyping of `HeapType`.

__TODO__: Revisit all error generation for more accurate validation error messages and proper treatment of bad casts (as a result of incorrect indexing into heap type table).


src/engine/CodeValidator.v3
```diff
@@ -1087,6 +1087,53 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
 				I64_ATOMIC_RMW_CMPXCHG	=> check_MEMARG(opcode, 8);
 				I32_ATOMIC_RMW8_CMPXCHG_U	=> check_MEMARG(opcode, 1);
 				I32_ATOMIC_RMW16_CMPXCHG_U	=> check_MEMARG(opcode, 2);
+				CONT_NEW => {
+					var cont_decl = parser.readCont();
```
titzer, line 1091: Should check for a null cont_decl here.

src/engine/BinParser.v3
```diff
@@ -421,7 +421,9 @@ class BinParser(extensions: Extension.set, limits: Limits, filename: string) ext
 	def readDefContType(final: bool, supertypes: Array<HeapType>) {
 		var sig_index = parser.readU32_i("cont sigature index", u32.!(module.heaptypes.length));
 		if (!err.ok()) return;
-		var decl = ContDecl.new(final, supertypes, u32.!(sig_index));
+		var sig_htype = module.heaptypes[sig_index];
+		// TODO[ss]: Safe casting (throw error on failure)
```
titzer, line 425: This should check the type of the declaration and signal an error here.

src/engine/CodeValidator.v3
```diff
@@ -1087,6 +1087,53 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
...
+					popE(ValueTypes.Ref(true, cont_decl.sig));
+					// TODO[ss]: nullable = false as per proposal doc. Confirm/ask this.
+					val_stack.push(ValueTypes.Ref(false, cont_decl));
```
titzer, line 1094: I think that's right here; this should produce a non-null reference to a continuation.

src/engine/CodeValidator.v3
```diff
@@ -1087,6 +1087,53 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
...
+				}
+				CONT_BIND => {
+					var cont_a = parser.readCont();
```
titzer, line 1097: Need to check for null here too.

src/engine/CodeValidator.v3
```diff
@@ -1087,6 +1087,53 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
...
+					popE(ValueTypes.Ref(true, cont_a));
+					var cont_b = parser.readCont();
+
+					var sig_a = cont_a.sig;
```
titzer, line 1101: Maybe `input_sig` and `output_sig`?

src/engine/CodeValidator.v3
```diff
@@ -1087,6 +1087,53 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
...
+					}
+
+					var n_binds = source_length - target_length;
+					var applied_args = Arrays.range(sig_a.params, 0, n_binds);
+					var remain_args = Arrays.range(sig_a.params, n_binds, source_length);
+
+					// Resulting signature after partial application matches the given signature `sig_b`.
+					// TODO[ss]: Make sure subtyping is checked correctly here.
+					var result_sig = SigDecl.new(true, ValueTypes.NO_HEAPTYPES, remain_args, sig_a.results);
+					if (!result_sig.isAssignableSig(sig_b)) {
```
titzer, line 1117: We should maybe factor out a different routine that doesn't work on sigs, but parameters and returns (or maybe just an Array<ValueType>, because the `isAssignableSig` is a hold-over from a more general structural type system that was not adopted in Wasm.

src/engine/CodeValidator.v3
```diff
@@ -1087,6 +1087,53 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
...
+						err_atpc().IllegalContinuationBind(cont_a, cont_b);
+					}
+
+					// Values on the stack matches the consumed arguments `applied_args`.
+					checkAndPopArgs(applied_args);
+					val_stack.push(ValueTypes.Ref(false, cont_b));
+				}
+				SUSPEND => {
+					var tag = parser.readTagRef();
```
titzer, line 1126: Have to check for null here, in case the index was OOB or not a tag.

src/engine/CodeValidator.v3
```diff
@@ -1266,6 +1351,15 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
 			if (!ValueTypes.isAssignable(ct[j], p[j])) err_atpc().TypeMismatchInCatch(i, p[i], ct[i]);
 		}
 	}
+	def checkContHandle(i: int, tag: TagDecl, target: ControlEntry) {
+		var p = labelArgs(target);
+		if (p == null) return;
+		var ct = Arrays.append(ValueType.Ref(true, HeapType.ANY), tag.fields);
```
titzer, line 1357: Is `HeapType.ANY` right here? I think this may be a `ref cont`?

src/engine/CodeValidator.v3
```diff
@@ -1266,6 +1351,15 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
...
+		if (ct.length != p.length) return err_atpc().ArityMismatchInCatch(i, p.length, ct.length);
+		for (j < p.length) {
+			if (!ValueTypes.isAssignable(p[j], ct[j])) err_atpc().TypeMismatchInCatch(i, p[i], ct[i]);
```
titzer, line 1360: Let's update the error method to `TypeMismatchInHandler` and have it take an argument for "catch" (i.e. catch handler) and "resume" (i.e. resume handler).

src/engine/CodeValidator.v3
davidmaamoaix: Sorry I forgot to ask about this. Currently it seems that (I'm not sure if this is really the case, but this behavior happened in the test cases I'm using) when reading a `(ref $cont)` as the result of a block, the parser interprets it as a `Ref(..., HeapType.ANY)`. I'm guessing this might be because the bytecode format of `ref` is a generic index into `heaptypes`. This results in the top value in `labelArgs(block)` is a `ref any`. I used `Ref(true, HeapType.ANY)` as a placeholder.

davidmaamoaix: Followup: this seems to be caused by `readHeapType` returning `ANY` on reading `ref cont`. I must've missed this in my previous PR. Will fix this.

davidmaamoaix:
The mentioned issues should be addressed now. Thanks!.

src/engine/CodeValidator.v3
```diff
@@ -1087,6 +1087,91 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
...
+					popE(ValueTypes.Ref(true, ct));
+
+					var thrown_tag = parser.readTagRef();
+					if (thrown_tag == null) return;
+					checkAndPopArgs(thrown_tag.fields);
+
+					var handlers = parser.readContHandlers();
+					for (i < handlers.length) {
+						var handler = handlers[i];
+						var target = getControl(handler.depth);
```
titzer, line 1168: I just had another look at the rules, and I think the label expects a new continuation type to be constructed out of input continuation type somehow. It's a little subtle, and problematic with canonicalization. Given that, I think we can land this and iterate on it. Might need to discuss with WasmFX people via an issue.

titzer (approved):

PR #162, opened 2024-05-22
[typed-cont] Stricter check for continuation signature in cont.resume

davidmaamoaix:
This PR adds continuation type checking when validation the handler table for `cont.resume` and `cont.resume_throw`.

### DIFF
--- a/src/engine/CodeValidator.v3
+++ b/src/engine/CodeValidator.v3
@@ -1350,10 +1350,47 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
 	def checkContHandle(i: int, tag: TagDecl, cont: ContDecl, target: ControlEntry) {
 		var p = labelArgs(target);
 		if (p == null) return;
-		var ct = Arrays.append(ValueType.Ref(true, HeapType.Cont(cont)), tag.fields);
-		if (ct.length != p.length) return err_atpc().ArityMismatchInHandler("resume", i, p.length, ct.length);
-		for (j < p.length) {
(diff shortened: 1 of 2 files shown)


src/engine/CodeValidator.v3
```diff
@@ -1351,10 +1337,20 @@ class CodeValidator(extensions: Extension.set, limits: Limits, module: Module, e
...
+			if (target == null) return;
+			checkContHandle(i, handler.tag, cont, target);
+			ctlxfer.refC(target, handler.tag, false);
+		}
+	}
+	def checkContHandle(i: int, tag: TagDecl, cont: ContDecl, target: ControlEntry) {
 		var p = labelArgs(target);
 		if (p == null) return;
-		var ct = Arrays.append(ValueType.Ref(true, HeapType.Cont(null)), tag.fields);
+		var ct = Arrays.append(ValueType.Ref(true, HeapType.Cont(cont)), tag.fields);
```
titzer, line 1353: My reading of the rule for `resume` is that the parameter types of the continuation expected at the label are the results from the tag, and the result types of the continuation expected at the label are the same (or subtypes) of the result of the original `cont`.

davidmaamoaix:
My bad. I also used the wrong `.bin.wast` program for testing during the last commit, which passed despite the type check issue ;(

I updated `checkContHandle` to reflect the changes so that the continuation ref argument expected by the label is also type checked properly. I'm using structural subtyping checks on their signature instead of `ValueTypes.isAssignable` as per the discussion [here](https://github.com/WebAssembly/stack-switching/issues/57). Please let me know if this is appropriate or if we need to change that behavior.

titzer (approved):
Ok, let's land this and let the test suite do its thing :-)

PR #163, opened 2024-05-29
Wavid fixing wizeng

wavid-b:


wavid-b:
Adds an addition check before returning it an error

titzer (approved):

PR #164, opened 2024-05-30
made "-csv" option work for hotness monitor

wavid-b:
also minor code cleanup for readability


wavid-b:
Made more updated verison - pull request on that one instead

PR #165, opened 2024-05-30
Added -csv functionality for hotness monitor

wavid-b:
-minor code cleanup 
-different from old pull with same name; creates 1 buffer then pushes all at once for reduced complexity


wavid-b:
Closed because new PR has same functionality with better syntax

PR #166, opened 2024-05-30
Loop and Hotness monitor CSV functionality 

wavid-b:


src/monitors/HotnessMonitor.v3
```diff
@@ -162,15 +180,25 @@ private class HotnessData(m: Module) {
...
+		OUT.puts("func,pc,count\n");
+		for (i < result.length) {
+			var r = result[i];
+			//add each line of data so the out buffer
+			OUT.put1("%q", r.func.render(m.names, _))
+				.putc(',')
+				.put1("+%d", r.pc)
+				.putc(',')
+				.putd(r.count);
+			if(i!=result.length-1){
```
titzer, line 196: Formatting nit: spaces between operators here.

src/monitors/LoopMonitor.v3
```diff
@@ -183,3 +188,21 @@ def report(entries: Vector<LmEntry>) {
...
 }
+def reportCsv(entries: Vector<LmEntry>) {
+	OUT.puts("func,pc,end_pc,count\n");
+	for (i < entries.length) {
+		var e = entries[i];
+		if (e.count == 0 && !showZeroCounts) continue;
+		//add each line of data so the out buffer
+		OUT.put1("%q", e.func.render(e.module.names, _))
+			.putc(',')
+			.putd(e.pc)
```
titzer, line 199: Whether there is a `+` before `pc` should be consistent between all the monitors; I vote to not have it, since it's a numerical quantity meant to be consumed by machines.

doc/Instrumentation.md
```diff
@@ -23,15 +23,15 @@ wizeng --monitors=loops,coverage{code=instr} test.wasm
 
 ## Options
 
-When no optional arguments are provided, most monitors display only summaries. The `code` argument will print out instrumentation infromation alongside bytecode, where possible. The available monitors and their options are as follows:
+When no optional arguments are provided, most monitors display only summaries. The `code` argument will print out instrumentation infromation alongside bytecode, where possible. The `csv` argument outputs the monitor data in a csv format to the terminal. The available monitors and their options are as follows:
```
titzer, line 26: I would say "to standard out" rather than "to the terminal".

doc/Instrumentation.md
```diff
@@ -23,15 +23,15 @@ wizeng --monitors=loops,coverage{code=instr} test.wasm
...
 
 | Monitor    | Options                                    | Description                                                                                         |
 |------------|--------------------------------------------|-----------------------------------------------------------------------------------------------------|
-| `loops`    | `{c\|code}`                                | Analyzes loop blocks and their execution behaviour.                                                 |
+| `loops`    | `{c\|code}` `{csv}`                        | Analyzes loop blocks and their execution behaviour.                                                 |
```
titzer, line 30: The `-csv` option is external to this, so this might be a misleading. I've been meaning to think about cleaning up how all the options are parsed and now might be a good time to think about a good strategy for all the monitors' options.

src/monitors/LoopMonitor.v3
wavid-b: I'll change the ones I've interacted with on the csv ouputs

src/monitors/HotnessMonitor.v3
wavid-b: will do

PR #169, opened 2024-06-03
[typed-continuation] V3 interpreter execution

davidmaamoaix:
This PR implements the interpreter execution of new opcodes introduced in the typed-continuation proposal.


build
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash
```
titzer, line 1: Seems like a spurious copy of `build.sh` here?

src/engine/Execute.v3
```diff
@@ -32,10 +32,12 @@ component Execute {
 	}
 }
 
+type SuspendInfo(tag: Tag, vals: Array<Value>) #unboxed;
+
 type Result {
 	case Value(vals: Array<Value>) {}
 	case Throw(thrown: Throwable) {}
-	case Suspend(stack: WasmStack, exntag: Tag, valrefs: Range<Value>);
+	case Suspend(stack: WasmStack, info: SuspendInfo);
```
titzer, line 40: I'm wondering if the combination of `(stack, info)` should just be a continuation reference. wdyt?

davidmaamoaix: I think this is a good replacement for `stack`, although I'm not sure how to handle the returned `tag` and `vals` by the suspend instruction. I'm thinking since after executing something like `resume $ct` where `$ct : [t1*] -> [t2*]`, `$ct` will need to suspend with a value array of `[t2*]` which is then pushed onto the parent stack. I'm wondering if this value array should be kept separate from the continuation reference as it is part of the information associated with that specific `suspend` rather than a part of the returned continuation itself.

titzer: Oh, yes, I think you're right.

davidmaamoaix:
The new opcodes should be supported by the interpreter now. I'll look into setting up the spectests (and fixing the tag import format compatibility).

src/engine/v3/V3Interpreter.v3
```diff
@@ -1108,6 +1109,91 @@ class V3Interpreter extends WasmStack {
...
+				var out_cont = ContDecl.!(frame.func.instance.heaptypes[codeptr.read_uleb32()]);
+				var n_binds = in_cont.sig.params.length - out_cont.sig.params.length;
+
+				var cont = Continuation.!(Value.Ref.!(pop()).val);
+				if (cont == null) trap(TrapReason.NULL_DEREF);
+				// continuation can only be resumed/bound once
+				if (cont.used) trap(TrapReason.USED_CONTINUATION);
+
+				var new_cont = Continuation.new(cont.cont, cont.stack);
+				for (i < cont.bounded.length) new_cont.bounded.put(cont.bounded[i]);
```
titzer, line 1134: I don't see a use of these bound parameters. If they are prefix of the continuation's expected arguments, I think you should be able to just push them onto the continuation's stack?

src/engine/v3/V3Interpreter.v3
```diff
@@ -1108,6 +1109,91 @@ class V3Interpreter extends WasmStack {
...
+				var cont_type = ContDecl.!(frame.func.instance.heaptypes[cont_idx]);
+				var handlers = codeptr.read_handlers();
+				
+				var cont = Continuation.!(Value.Ref.!(pop()).val);
+				if (cont == null) trap(TrapReason.NULL_DEREF);
+				if (cont.used) trap(TrapReason.USED_CONTINUATION);
+
+				cont.used = true;
+				cont.stack.bind(values.peekn(cont_type.sig.params.length));
+				var r = cont.stack.resume();
```
titzer, line 1153: This works for now, but leaves the Virgil frames of the calling stack on the Virgil stack. That means that a long chain of resumes could overflow the Virgil stack. Let's think about how to fix this. I think we'll need to refactor the loop that calls `runWasmCode` so that there's a guard around it to find handlers for suspends and switch to the right stack.

src/engine/v3/V3Interpreter.v3
```diff
@@ -1108,6 +1109,91 @@ class V3Interpreter extends WasmStack {
...
+				var range = values.peekn(n_binds);
+				for (i < range.length) new_cont.bounded.put(range[i]);
+				push(Value.Ref(cont));
+
+				values.top -= n_binds;
+			}
+			RESUME => {
+				var cont_idx = codeptr.read_uleb32();
+				var cont_type = ContDecl.!(frame.func.instance.heaptypes[cont_idx]);
+				var handlers = codeptr.read_handlers();
```
titzer, line 1145: This is fine for now, but I think we'll want to use the sidetable entry to directly get to the handler table. For example, the fast interpreter should avoid decoding the handlers and instead have a direct pointer to a handler table in the sidetable, as well as the length of the `resume` opcode because it will fall through to the next opcode upon return.

src/engine/v3/V3Interpreter.v3
```diff
@@ -1108,6 +1109,91 @@ class V3Interpreter extends WasmStack {
...
+							suspend_info = SuspendInfo(info.tag, info.vals, parent);
+						}
+					}
+				}
+			}
+			RESUME_THROW => {
+				var cont_idx = codeptr.read_uleb32();
+				var cont_type = ContDecl.!(frame.func.instance.heaptypes[cont_idx]);
+
+				var exn_tag = frame.func.instance.tags[codeptr.read_uleb32()];
```
titzer, line 1172: Can you factor out all the expressions that are `codeptr.read_XXX` to the beginning of the handler case? This makes it easier to see what is read from the code stream.

src/engine/v3/V3Interpreter.v3
davidmaamoaix: Sorry I don't think I quite get the part of "having a direct pointer to a handler table". By this do you mean to cache/canonicalize the handler information somewhere else in the module during initialization, and iterate through the cached version to find the appropriate tag during `resume` (like the way `try_table` are implemented)?

src/engine/v3/V3Interpreter.v3
davidmaamoaix: Sorry I must've missed this. I think in our last meeting we discussed moving this part out of the WasmStack and manage the stack switching `V3InterpreterOnlyStrategy`. I will try to implement that.

src/engine/v3/V3Interpreter.v3
titzer: Right, similar to exception handling, there should be a handler table hanging off of the function that is searched when suspending. IIUC then the handler table for the top frame of a stack only needs to be searched upon suspend (no iteration of frames, just stacks via parent pointers). It might even be possible to use the exact same format of handler tables and share the search logic.

titzer (approved):

PR #172, opened 2024-07-03
Implementing Atomic Load and Store Bytecodes in Fast Interpreter

khagankhan:
## Atomic Load and Store Bytecode Implementations

Implements the following bytecodes in fast interpreter:

- `I32_ATOMIC_LOAD`
- `I64_ATOMIC_LOAD`
- `I32_ATOMIC_LOAD8_U`
- `I32_ATOMIC_LOAD16_U`
- `I64_ATOMIC_LOAD8_U`
- `I64_ATOMIC_LOAD16_U`
- `I64_ATOMIC_LOAD32_U`
- `I32_ATOMIC_STORE`
- `I64_ATOMIC_STORE`
- `I32_ATOMIC_STORE8`
- `I64_ATOMIC_STORE8`
- `I32_ATOMIC_STORE16`
- `I64_ATOMIC_STORE16`

Add regression tests for atomic store bytecodes.


src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -1302,6 +1303,39 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
 		bindHandler(Opcode.I64_STORE16);
 		genStore(asm.movw_m_r);
 	}
+	def genAtomicLoad(a: X86_64Gpr, b: X86_64Addr) -> this {
```
titzer, line 1306: This generates the same code for every atomic load and store, regardless of the size. There should be different instructions in for 8, 16, and 32 bit loads. Also, it looks like loads actually do store to memory via the `xchg`. For example, see the V8 implementation here: https://github.com/v8/v8/blob/main/src/wasm/baseline/x64/liftoff-assembler-x64-inl.h#L544

khagankhan: I see. It appears my assumption was wrong for xchg_m_r  for loading as well. I will address them right away.

PR #173, opened 2024-07-09
custom monitor for bytecode performance within wasm module

thelongmarch-azx:
When executing a compiled language (C, rust etc.) in a wasm engine, we simply target wasm output in the source compiler, and can easily use existing profiling/instrumentation capabilities provided in popular wasm engine that monitor source code performance.

However, if we want to run interpreted language in a wasm engine, we typically need to first turn the source code into interpreter specific bytecode, then compile the source language interpreter into a single wasm module, along with the source bytecode. In this case, it is difficult to monitor/profile source code performance simply by monitoring compiled wasm functions, because we'd be observing the interpreter program's behavior, instead of the source program behavior.

This PR implements a new type of monitor that removes this layer of indirection, by using the instrumentation capabilities in the wizard engine. It is able to recover source code execution, attribute accurate fuel count for each "step" of source bytecode execution, recover source level function call traces and recover hot paths taken by the source code.

The implementation assumes a "switch over loop" interpreter design, ad uses static analysis to estimate the dispatch function containing the interpreter loop, then monitor for byte level memory read to reconstruct the source level bytecode consumed by the interpreter.


src/monitors/BytecodeProfilingMonitor.v3
```diff
@@ -0,0 +1,450 @@
...
+						b.put3("%x,%d,%d", i, j, bytecode_hotness[i][j]).ln();
+					costs = costs.tail;
+				}
+			}
+		}
+		System.write(fd, b.toString());
+	}
+}
+
+private class CombinedCountProbe(all_counters: List<DynamicLoc -> Resumption>) extends CountProbe {
```
titzer, line 408: I wasn't quite sure why we need a combined counter; AFAICT this is just a count of all the `br_table` executions?

src/monitors/BytecodeProfilingMonitor.v3
```diff
@@ -0,0 +1,450 @@
...
+// as dispatch.
+var MIN_DISPATCH_CASE_COUNT = 100;
+
+// the maximum size in bytes that separates 2 consecutive opcodes.
+var MAX_OPERAND_LEN = 6;
+
+class BytecodeProfilingMonitor extends Monitor {
+	// all functions that could be the dispatch function.
+	// kvp of function index to the (FuncDecl, bytecode_read_pc, profile_state) pair.
+	var potential_dispatch_functions = HashMap<int, (FuncDecl, int, SourceProfileState)>.new(int.!<int>, int.==);
```
titzer, line 26: I think you could replace this with just a list, and have each probe have a reference to its `SourceProfileState`, to avoid a lookup when probes fire.

src/monitors/BytecodeProfilingMonitor.v3
```diff
@@ -0,0 +1,450 @@
...
+			if (f.imp != null) continue;
+			bi = bi.reset(f);
+
+			// We identify dispatch function through static analysis, by looking for:
+			// 	1. a sequence of consequtive BLOCK instructions exceeding MIN_DISPATCH_CASE_COUNT.
+			// 	   This corresponds to a switch statement over all possible opcode in the source code.
+			// 	2. a I32_LOAD8_U instruction, followed by a BR_TABLE instruction, in the inner scope of the
+			// 	   nested BLOCK chain.
+			var prev_OPCODE_READ_pc = -1, bytecode_read_pc = -1;
+			var last_block_chain = 0;
```
titzer, line 61: What about instead just matching on the number of cases in a `br_table`? That way you don't have to track the number of consecutive blocks, but it'd rather be determined by how many cases there are (though a `br_table` might have many redundant entries).

titzer (commented):
Some initial comments.

src/monitors/BytecodeProfilingMonitor.v3
thelongmarch-azx: Good idea, will make the change

src/monitors/BytecodeProfilingMonitor.v3
```diff
@@ -0,0 +1,423 @@
+// Copyright 2023 Wizard Authors. All rights reserved.
```
titzer, line 1: 2024

src/monitors/BytecodeProfilingMonitor.v3
```diff
@@ -0,0 +1,423 @@
...
+
+// the maximum size in bytes that separates 2 consecutive opcodes.
+var MAX_OPERAND_LEN = 6;
+
+class BytecodeProfilingMonitor extends Monitor {
+	def visitor = DispatchFunctionBytecodeVisitor.new(BytecodeIterator.new());
+	def fuel_count_probe = FuelCountProbe.new();
+	// the SourceProfileState of all functions that could be the dispatch function.
+	var potential_dispatch_functions: List<SourceProfileState>;
+	var all_counters: List<DynamicLoc -> Resumption>;
```
titzer, line 28: Looks like this field is no longer used.

src/monitors/BytecodeProfilingMonitor.v3
```diff
@@ -0,0 +1,423 @@
...
+			states.head.hotness.report();
+		}
+	}
+
+	private def instrumentFunc(mi: ModuleInstrumenter, func: FuncDecl) {
+		visitor.reset();
+		visitor.bi.reset(func).dispatchAll(visitor);
+		if (visitor.is_dispatch_func) { // save the profile state and insert memory probes 
+			var state = SourceProfileState.new(fuel_count_probe, func.func_index,
+				StringBuilder.new().put1("func_%d", func.func_index).toString());
```
titzer, line 73: You can get the function name by calling `.render` and passing the name section from the module.

src/monitors/BytecodeProfilingMonitor.v3
```diff
@@ -0,0 +1,423 @@
...
+			
+			potential_dispatch_functions = List.new(state, potential_dispatch_functions);
+			var memory_visitor = MemoryVisitor.new(mi.module,
+				beforeMemRead(state, visitor.opcode_read_pc, _, _, _, _), null, null);
+			memory_visitor.bi = visitor.bi.reset(func);
+			memory_visitor.bi.dispatchAll(memory_visitor);
+		}
+	}
+
+	private def beforeMemRead(profile_state: SourceProfileState, opcode_read_pc: int,
```
titzer, line 83: It should be possible to separate the opcode read from other reads statically based on the pc and add a different probe at different locations?

src/monitors/BytecodeProfilingMonitor.v3
```diff
@@ -0,0 +1,423 @@
...
+	// updates profiler state upon wasm function invocation/return, during the execution of a
+	// valid source opcode. This is used to understand which wasm functions are associated to
+	// particular opcodes.
+	def wasmFuncEvent(wasm_func_id: int, is_start: bool) {
+		var wasm_trace = WasmFuncTrace(wasm_func_id, u32.!(count_probe.count) - last_step_count, is_start);
+		cur_wasm_func_trace = List.new(wasm_trace, cur_wasm_func_trace);
+	}
+
+	// updates the profiler state upon source function invocation.
+	def sourceFuncStart() {
```
titzer, line 162: We should pick a good name to refer to the higher-level functions and bytecode. What about "guest function" and "guest bytecode"?

src/monitors/BytecodeProfilingMonitor.v3
thelongmarch-azx: definitive proof that software engineering is just strategic copy/paste-ing

src/monitors/BytecodeProfilingMonitor.v3
thelongmarch-azx: done

src/monitors/BytecodeProfilingMonitor.v3
thelongmarch-azx: done

src/monitors/BytecodeProfilingMonitor.v3
thelongmarch-azx: I made them separate by using 2 MemoryVisitors, saving the outermost conditional branching

PR #174, opened 2024-07-16
Added LazyImportOption

MilesConn:


src/engine/Instantiator.v3
```diff
@@ -14,8 +14,10 @@ class ImportProcessor(name: string) {
 	def postprocess(error: ErrorGen, instance: Instance) { }
 }
 
+class ImportOptions(allowLazyImports: bool) {}
+
 // Creates {Instance} objects, given a module and a list of imports.
-class Instantiator(extensions: Extension.set, module: Module, var imports: Array<Exportable>, error: ErrorGen) { // TODO: s/error/err
+class Instantiator(extensions: Extension.set, module: Module, var imports: Array<Exportable>, error: ErrorGen, importOptions: ImportOptions) { // TODO: s/error/err
```
titzer, line 20: I think this can be made significantly simpler by adding an important processor which binds functions to be lazy if they haven't been bound. That way, it can just be added at the end by an option to the `wizeng.main`.

PR #175, opened 2024-07-18
[typed-cont] Stack switching implementation for fast interpreter

linxuanm:
This PR implements new opcodes introduced in stack switching proposal for the fast interpreter.

### DIFF
--- a/src/engine/v3/V3Interpreter.v3
+++ b/src/engine/v3/V3Interpreter.v3
@@ -1128,14 +1128,15 @@ class V3Interpreter extends WasmStack {
 			I8X16_SWIZZLE => doSwizzle();
 			// ext:typed-continuation
 			CONT_NEW => {
-				var cont_index = codeptr.read_uleb32();
-				var cont_decl = ContDecl.!(frame.func.instance.heaptypes[cont_index]);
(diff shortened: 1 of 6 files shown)


src/engine/v3/V3Interpreter.v3
```diff
@@ -1128,14 +1128,15 @@ class V3Interpreter extends WasmStack {
 			I8X16_SWIZZLE => doSwizzle();
 			// ext:typed-continuation
 			CONT_NEW => {
-				var cont_index = codeptr.read_uleb32();
-				var cont_decl = ContDecl.!(frame.func.instance.heaptypes[cont_index]);
-				var func = Function.!(Value.Ref.!(pop()).val);
-				if (func == null) trap(TrapReason.NULL_DEREF);
+				// var cont_index = codeptr.read_uleb32();
```
titzer, line 1131: Can you keep the reading of the `cont_index` a separate line here? That way it's easier to see the decoding of immediates, rather than that being nested in an argument below.

linxuanm:
This PR should be finished (implementation for `resume`, `cont.new` and `cont.bind`).

PR #176, opened 2024-07-24
Implement wasm R3 frontend in wizeng

thelongmarch-azx:
A new monitor that records wasm module execution traces, including interaction with host environment, and produce event traces compatible with the wasm R3 backend to produce standalone replay module


doehyunbaek:
Interesting work, could you tell me how to test out this work? Just simple bash commands would be enough I think.

src/monitors/R3Monitor.v3
```diff
@@ -0,0 +1,572 @@
...
+
+	def onParse(module: Module) {
+		handler = EventHandler.new(module);
+		Instrumentation.insertGlobalProbe(ReturnProbe.new(handler));
+		var mm = ModuleInstrumenter.new(module);
+		var visitor = R3MonitorBytecodeVisitor.new(handler, module, BytecodeIterator.new());
+		mm.forEachFunc(instrumentFunc(visitor, _, _));
+		mm.beforeMemRead(handler.memoryRead);
+		mm.beforeMemWrite(handler.memoryWrite);
+		mm.beforeMemGrow(handler.memoryGrow);
```
titzer, line 21: We'll also need to instrument `memory.init`, `memory.copy`, and `memory.fill`.

src/monitors/R3Monitor.v3
```diff
@@ -0,0 +1,572 @@
...
+		System.puts(sb.toString());
+	}
+
+	private def instrumentFunc(visitor: R3MonitorBytecodeVisitor, mm: ModuleInstrumenter, func: FuncDecl) {
+		Instrumentation.insertLocalProbe(mm.module, func.func_index, 0, EnterProbe.new(handler, func));
+		visitor.bi.reset(func).dispatchAll(visitor);
+	}
+}
+
+private class R3MonitorBytecodeVisitor(handler: EventHandler, module: Module, bi: BytecodeIterator) extends BytecodeVisitor {
```
titzer, line 40: You can save one method if this class extends `BytecodeInstrumenter`.

src/monitors/R3Monitor.v3
```diff
@@ -0,0 +1,572 @@
...
+		trace.put(WasmEvent.TableGrow(table_idx, growth));
+		shadowTables[table_idx].grow(growth, Values.FUNCREF_NULL);
+	}
+
+	def memoryRead(loc: DynamicLoc, mem: Memory, addr: u64, size: u64) -> Resumption {
+		var addr = u32.view(addr);
+		var size = u32.view(size);
+		var mem_index = mem.decl.memory_index;
+		var shadow_mem = shadowMems[mem_index];
+		var mem_ptr = mem.range_oil_32(addr, 0, size).result;
```
titzer, line 248: You can save a little here by using `range_ol_64`, as the index is `0`.

src/monitors/R3Monitor.v3
```diff
@@ -0,0 +1,572 @@
...
+		shadowTables[table_idx][entry_idx] = new_val;
+	}
+
+	def onTableGrow(table_idx: int, growth: u32) {
+		trace.put(WasmEvent.TableGrow(table_idx, growth));
+		shadowTables[table_idx].grow(growth, Values.FUNCREF_NULL);
+	}
+
+	def memoryRead(loc: DynamicLoc, mem: Memory, addr: u64, size: u64) -> Resumption {
+		var addr = u32.view(addr);
```
titzer, line 244: If you use the 64-bit memory range methods, you won't need to do this view, and it will be forward-compatible with Wasm memory64.

src/monitors/R3Monitor.v3
thelongmarch-azx: done

thelongmarch-azx:
> Interesting work, could you tell me how to test out this work? Just simple bash commands would be enough I think.

Yes, it will require some custom shell script. What I'm thinking at the moment is gathering a bunch of wasm modules that has a "_start" routine and imports wsp to communicate with the OS, and use the existing R3 frontend to generate a set of reference traces, then compare the output of this monitor with the ref traces. There are some small differences between the original R3 trace and my trace, so I'll probably need to massage it a bit.

doehyunbaek:
> Yes, it will require some custom shell script. 

If you are willing to share your custom shell script I am willing to adapt too!

> What I'm thinking at the moment is gathering a bunch of wasm modules that has a "_start" routine and imports wsp to communicate with the OS, and use the existing R3 frontend to generate a set of reference traces, then compare the output of this monitor with the ref traces. 

Seems like a solid plan.

> There are some small differences between the original R3 trace and my trace, so I'll probably need to massage it a bit.

I think I am going to refactor Wasm-R3 on October so if the change to your trace is reasonable, I am down to adapt. Don't think the original R3 trace is something unchangeable.

thelongmarch-azx:
> If you are willing to share your custom shell script I am willing to try!
I don't have a script ready yet. I want to keep all the tests within the wizard testing setup. That means I'll generate some reference trace files first, put them here (https://github.com/titzer/wizard-engine/tree/master/test/monitors/expected), and update this testing script (https://github.com/titzer/wizard-engine/blob/master/test/monitors/test.sh) to match the stdout from my monitor to the reference trace files.

In addition, due to the differences between my trace output and the original trace output, I just compare the 2 manually, and use my output trace as the reference. Furthermore, wasabi doesn't support simd, and a lot of complex WASI modules use SIMD, so I can't use the original r3 frontend to generate a trace. For those test cases, I still want to include them, so I use the existing r3 backend, with my trace to produce the combined module, then compare the memory read traces of the combined module with the original module.

> I think I am going to refactor Wasm-R3 on October so if the change to your trace is reasonable, I am down to adapt. Don't think the original R3 trace is something unchangeable.
I made my traces fully compatible with the current r3 backend. I just omitted some fields that aren't needed to generate the IR. Please see "type WasmEvent" in my monitor for details.

doehyunbaek:
Thanks for the detailed expalanations! Few comments:

> I made my traces fully compatible with the current r3 backend.

Yeah I don't think current r3 backend is going to change much, it is pretty stable, at least its parsing of trace format. I think it's good that your traces are fully compatible with the current r3 backend.

> In addition, due to the differences between my trace output and the original trace output, I just compare the 2 manually, and use my output trace as the reference.

At least for the 80+16 test cases checked in the Wasm-R3 repo, I think it is ideal that Wasm-R3 and Wizard agrees on the trace output, what do you think? If you want, I am willing to put (standalone replay.wasm, trace.r3) pair onto somewhere we can both access, maybe the directory you mentioned. They all don't use SIMD so existing Wasm-R3 frontend and wizard's new frontend should be able to generate the same trace.

doehyunbaek:
Also sola-st/wasm-r3's main branch is quite a mess now(CI has been broken for a long time), so I think it would be best to have our reference implementation of Wasm-R3 as a OOPSLA_2024 branch(https://github.com/sola-st/wasm-r3/tree/OOPSLA_2024) which I confirmed all 96 tests we mentioned passing.

thelongmarch-azx:
> Also sola-st/wasm-r3's main branch is quite a mess now(CI has been broken for a long time), so I think it would be best to have our reference implementation of Wasm-R3 as a OOPSLA_2024 branch(https://github.com/sola-st/wasm-r3/tree/OOPSLA_2024) which I confirmed all 96 tests we mentioned passing.

The main issue with the wizard monitor is that wizard engine doesn't have any linking/pre-loading capabilities, so we can't really test a "index" module that imports something custom from another wasm module, this really limits the types of tests we can use. We also cannot arbitrarily call exported functions of a module when running in wizard. When you run a module in wizard, only the "main" or "_start" function gets invoked.  

My email is anzixi0919@gmail.com, I think we can schedule a 30 min meeting to go over some details. I'm available 9am to 11am EST on Friday, any time after 8pm next Monday, or any time between 9am to midnight after next Monday.

doehyunbaek:
Sure. I'll send an invite to 9am EST friday.

> The main issue with the wizard monitor is that wizard engine doesn't have any linking/pre-loading capabilities, so we can't really test a "index" module that imports something custom from another wasm module, this really limits the types of tests we can use. 

I thought at least wizard can use standalone replay benchmark(replay.wasms) generated by Wasm-R3. We can talk over it.

If there's anything I need to know / prepare in advance, please let me know.

doehyunbaek:
We had a meeting and we agreed on these things.

1. It would be really beneficial to have same shared test cases that is runnable both on existing frontend and wizeng frontend.
2. To do so, wizeng frontend being able to "demarcate" the host functions in the standalone wasm benchmarks generated by Wasm-R3 would enable creation of the trace that exactly or roughly matches that of the existing trace.
3. There might be differences to the existing traces and the new ones, but they are not really big of a difference and can be easily matched.

I would share the (standalone replay benchmark, trace.r3) trace to same repo tomorrow. We can use that repo to coordinate our effort.

@titzer ping to let you know of our discussion.

titzer:
> We had a meeting and we agreed on these things.
> 
> 1. It would be really beneficial to have same shared test cases that is runnable both on existing frontend and wizeng frontend.

Yes, definitely.

> 2. To do so, wizeng frontend being able to "demarcate" the host functions in the standalone wasm benchmarks generated by Wasm-R3 would enable creation of the trace that exactly or roughly matches that of the existing trace.

Yeah, that sounds good. It should fall out naturally if the monitor can do a subset of functions.

> 3. There might be differences to the existing traces and the new ones, but they are not really big of a difference and can be easily matched.
> 
> I would share 96 (standalone replay benchmark, trace.r3) pairs to some repo tomorrow. We can use that repo to coordinate our effort.

Great!

Another thing I mentioned to Tom over slack is that we could craft a simple, custom host environment (a `HostModule` in Wizard, a JS script for wasm-r3) and then have a few tests that do an interaction with the host where we completely control the outcome, thus can create correct assertions.

doehyunbaek:
> we could craft a simple, custom host environment

Would the creation be automatic procedure? I need a little more deatil to see if your suggestion meets the requirements that I believe to be critical, which are:

- The test case should be same or mostly same(which will be the case for use I think, wizeng running statically linked standalone replay wasm and current wasm-r3 running dynamically linked wasm binaries)
- They should share a same backend(which in our case is a Wasm-R3's existing backend)
- The testing process should be fully automatic.

I think the approach we discussed satisfy all these goals.

doehyunbaek:
Here it is: https://github.com/doehyunbaek/wasm-r3-tests

thelongmarch-azx:
> > we could craft a simple, custom host environment
> 
> Would the creation be automatic procedure? I need a little more detail to see if your suggestion meets the requirements that I believe to be critical, which are:
> 
> * The test case should be same or mostly same(which will be the case for use I think, wizeng running statically linked standalone replay wasm and current wasm-r3 running dynamically linked wasm binaries)
> * They should share a same backend(which in our case is a Wasm-R3's existing backend)
> * The testing process should be fully automatic.
> 
> I think the approach we discussed satisfy all these goals.

This is just meant to be a quick sanity check, we'll still use the full test suite from your shared repo eventually.

doehyunbaek:
Ah, I see. I'm all for any stopgap testing mechanisms! Whatever works, is good.

doehyunbaek:
While reeimplementing Wasabi based frontend, I have noticed some apparent bugs, for example,

https://github.com/sola-st/wasm-r3/blob/1f75d09450d0d50ac87dcb70c3e66adc8002d191/src/tracer.cts#L384C1-L384C13

this always mutate shadowtable's 0th element, which is obviously not the intended behavior.

So for wasm-r3-tests that have table in their directory name, I think we reallly can't trust their reference output. I think we have to reason about them ourselves and decide which output is correct.

I think this place is a good place to discuss about any potential engineering decisions, or should I make a mailing thread about this topic?

thelongmarch-azx:
> While reeimplementing Wasabi based frontend, I have noticed some apparent bugs, for example,
> 
> https://github.com/sola-st/wasm-r3/blob/1f75d09450d0d50ac87dcb70c3e66adc8002d191/src/tracer.cts#L384C1-L384C13
> 
> this always mutate shadowtable's 0th element, which is obviously not the intended behavior.
> 
> So for wasm-r3-tests that have table in their directory name, I think we reallly can't trust their reference output. I think we have to reason about them ourselves and decide which output is correct.
> 
> I think this place is a good place to discuss about any potential engineering decisions, or should I make a mailing thread about this topic?

let's start a mailing thread. I also just added but 1 test into this monitor (I removed fib because the monitor takes forever to run to completion). I've updated my monitor with a "host_prefix" arg, where it will treat any wasm function with the given prefix as host function, thus enabling us to generate comparable traces while only running the replay module. However, the replay module changes the original function/global indices, so I used a python script to check the traces, instead of trying to match them directly. I also manually updated all the "L" trace events in the reference file, because the current tracer.cts prints them 1 byte at a time, whereas the wizard monitor does batch processing. I manually inspected each file I changed to make sure the values match.

test/monitors/wasm-r3-tests/validate-trace.py
thelongmarch-azx, line 1: @doehyunbaek please double check this script makes sense. I think we discussed these differences during our meeting last week, but feel free to ask any questions if I missed anything.

doehyunbaek: Sure!

doehyunbaek:
Thanks for the update! Excited to try them out.

I'm currently traveling in Taiwan for a vacation so I might be a little slow to catch up but I will do eventually. 

I wil make the email thread.

src/monitors/R3Monitor.v3
```diff
@@ -0,0 +1,728 @@
...
+			sb.puts("NaN");
+		} else if (val == Values.F32_infinity || val == Values.F64_infinity) {
+			sb.puts("Infinity");
+		} else if (val == Values.F32_minus_infinity || val == Values.F64_minus_infinity) {
+			sb.puts("-Infinity");
+		} else {
+			match (val) {
+				I32(val) => sb.put1("%d", i32.view(val));
+				I64(val) => sb.put1("%d", i64.view(val));
+				F32(val) => sb.put1("%s", Floats.displayFromBits32(val));
```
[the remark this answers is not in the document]
thelongmarch-azx: I prefer using something like "0x<hex_from_float>" in the trace to represent floats, it produces the most accurate representation. @doehyunbaek it should be easy to implement in the backend, right? we can also support both formats.

For this PR, I'll go back to using the hex format then, and modify the tests accordingly.

src/util/Floats.v3
```diff
@@ -44,4 +44,38 @@ component Floats {
 		if ((bits << 12) == 0) return false;
 		return true;
 	}
+	def displayFromBits32(bits: u32) -> string {
```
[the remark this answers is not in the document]
thelongmarch-azx: I removed these in favor of using hex representations for floats in the traces

src/monitors/R3Monitor.v3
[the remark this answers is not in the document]
doehyunbaek: I think it wouldnt be hard to support hex format in the backend.

Not sure Wasabi based frontend can output in hex format, let's go for each outputting its preferred format(Wasabi rounding, Wizard hex) and later decide if we can ditch the rounding format of current Wasabi implementation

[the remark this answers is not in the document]
titzer: Wasabi is probably using JS to render the value. But AFAICT there is no need to involve floating point types for memory, only for globals. It's best to avoid anything that doesn't preserve the exact floating point bits, so hex seems like the only solution going forward.

For the Wasabi frontend, one way to get the exact bits for globals is to would be to use a JS typed array. But that risks going through a JS value. Maybe better to have a Wasm module that imports a floating-point typed global, reads it, reinterprets it, and returns it as an `i64`.

titzer (approved):
lgtm, let's land this and iterate further in followups.

PR #182, opened 2024-08-12
Add opcodes for fast interpreter

khagankhan:
Add thread-related opcodes onto fast interpreter.


test/regress.failures
```diff
@@ -4,3 +4,22 @@ test/regress/ext:gc/array_new_fixed.bin.wast
 test/regress/ext:gc/br_on_non2.bin.wast
 test/regress/ext:gc/br_on_non.bin.wast
 test/regress/ext:gc/ref_eq.bin.wast
+test/regress/ext:threads/atomic_load_i32.bin.wast
```
titzer, line 7: Do these tests only fail on the single-pass compiler now? If so, the failures file should be at least platform-specific, since these should all pass on the V3 interpreter.

src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -1521,6 +1550,186 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
...
+		bindHandler(Opcode.I64_ATOMIC_RMW_CMPXCHG);
+		genAtomicCompareAndExchange(asm.q.cmpxchg_m_r);
+		bindHandler(Opcode.I32_ATOMIC_RMW8_CMPXCHG_U);
+		bindHandler(Opcode.I64_ATOMIC_RMW8_CMPXCHG_U);
+		genAtomicCompareAndExchange(asm.cmpxchgb_m_r);
+		bindHandler(Opcode.I32_ATOMIC_RMW16_CMPXCHG_U);
+		bindHandler(Opcode.I64_ATOMIC_RMW16_CMPXCHG_U);
+		genAtomicCompareAndExchange(asm.cmpxchgw_m_r);
+    }
+    def genAtomicOp<T>(op: (X86_64Addr, X86_64Gpr) -> T,
```
titzer, line 1639: This method is getting a little big. In particular, it takes a `bool` that switches off a lot of the functionality. I think it makes sense to split it into two methods at least, and factor out some of the commonality to shared functions.

test/regress.failures
khagankhan: Oh, it passes on V3 interpreter. I was planning to remove all of these when making a final PR to V3 interpreter.

src/engine/x86-64/X86_64Interpreter.v3
khagankhan: Sure thing! I didn’t like it either tbh. But since it was done in a single function I kept like that to get feedback

test/regress.failures
titzer: Ok, sgtm.

src/engine/x86-64/X86_64Interpreter.v3
titzer: Ok, I am going to merge this and we can fix it in a followup.

khagankhan: Hey! Thanks! I am going to make another PR about JIT and Virgil interpreters. I will write separate functions when making a PR about wait and notify o7

PR #185, opened 2024-09-06
R3 replay monitor (backend)

thelongmarch-azx:
This PR implements the wasm R3 backend in wizard engine. The new monitor parses a valid r3 trace and constructs the necessary import call contexts, then emulates the import function calls in the original module with these trace contexts.


src/engine/Value.v3
```diff
@@ -199,4 +199,14 @@ component Values {
 		}
 		return (null, -1);
 	}
+	def kind(val: Value) -> ValueKind {
```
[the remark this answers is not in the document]
thelongmarch-azx: the `toTag` method in spc takes a `ValueType` instead of a `Value` as an argument, so I still need to keep this method. However, there's another `ValueTypes.kind` function that does the same thing, I'll update spc to use that function

src/monitors/R3BackEndMonitor.v3
```diff
@@ -0,0 +1,165 @@
...
+
+class R3BackEndMonitor extends Monitor {
+    private var parser: R3TraceParser;
+    private var replay: HashMap<int, ReplayFunction>;
+    private var is_replay_valid = true;
+    private var mod: Module;
+    private var ins: Instance;
+    private var next_replay_context: Array<int>;
+
+    def configure(args: string) -> string {
```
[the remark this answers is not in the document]
thelongmarch-azx: sure, I'll try it out

src/monitors/MonitorOptions.v3
```diff
@@ -57,13 +57,32 @@ component MonitorOptions {
...
+				',' => {
+					arg = Arrays.prepend('-', Arrays.range(args, s, pos));
+					entry.o.parseOption(arg);
+					s = ++pos;
+				}
+				_ => pos++;
+			}
+		}
+		if (s < args.length) {
+			arg =Arrays.prepend('-', Arrays.range(args, s, args.length));
```
thelongmarch-azx, line 84: @titzer this, along with the changes in MonitorRegistry enables the monitors to effectively use `OptionGroup`. Basically each `MonitoryRegistryEntry` is now an `OptionGroup`, and as long as the options are specified in the individual monitors file, help text is automatically generated, and parsing happens automatically.

I intentionally made parsing happen before calling `configure` so that monitors can implement error handling logic

thelongmarch-azx: btw the help text looks like this now
Screenshot 2024-09-10 at 8 06 44 PM

titzer: As discussed in person, just a minor refactoring to use composition rather than inheritance here.

src/util/OptionsRegistry.v3
```diff
@@ -89,11 +89,11 @@ class OptionGroup(name: string, o: BasicOptions) {
 	def parseEmpty(str: string) -> bool {
 		return str == null;
 	}
-	def print(buf: StringBuilder) {
+	def print(buf: StringBuilder, indent: string) {
```
titzer, line 92: Since this already indents a little, maybe take the indent as an int here?

titzer (approved):
lgtm modulo the minor inheritance thing!

src/util/OptionsRegistry.v3
thelongmarch-azx: I ended up making this a flag `is_monitor_opt` and generates the monitor options help text exactly as expected

titzer (approved):

PR #188, opened 2024-09-11
ErrorGen case for errors in Whamm! probes

matthew-mojira:
ErrorGen for cases in which a probe function cannot be added in the Whamm monitor (e.g. type error at match time)


titzer (approved):
lgtm

PR #203, opened 2024-10-03
Move `stackDelta` to its own utility file 

matthew-mojira:
Need to ensure the `WasmParser` passed in is a copy and not one taken from the `CodeValidator`


src/util/StackDelta.v3
```diff
@@ -0,0 +1,167 @@
+component StackDelta {
```
titzer, line 1: How about moving this into `Opcodes`?

matthew-mojira: done

titzer (approved):

PR #206, opened 2024-10-08
Dynamic linking

thelongmarch-azx:
This feature enables bi-directional, dynamic linking of several wasm modules in wizard.


src/modules/ModuleLinker.v3
```diff
@@ -0,0 +1,233 @@
...
+// See LICENSE for details of Apache 2.0 license.
+
+/// Links the main module with the pre-loaded modules, and returns the main module instance.
+///
+/// The linker partially instantiates all the modules, then resolve each import to its original
+/// provider, then completes the instantiation. This allows bi-directional imports among multiple modules,
+/// as long as the import chain is acyclic.
+class ModuleLinker(engine: Engine,
+				   program_args: Array<string>,
+				   names: Array<string>,
```
titzer, line 11: It'd be a little more robust if these two arguments were an array of tuples, rather than parallel arrays of the same length.

src/wizeng.main.v3
```diff
@@ -39,82 +39,77 @@ def main(args: Array<string>) -> int {
...
-	var module: Module;
-	match (result) {
-		Ok(m) => {
-			module = m;
-			BasicTracing.instrumentModule(m);
-		}
-		FileNotFound => return ErrorBuilder.new()
-			.puts("wizeng: could not load file ")
+	var all_paths = Vector<string>.new();
+	if (WizengOptions.LINK_MODULES.val != null) {
```
titzer, line 43: As mentioned in person, some adjustments for putting the modules all on the command line are needed here.

src/wizeng.main.v3
```diff
@@ -39,82 +39,77 @@ def main(args: Array<string>) -> int {
...
+			ParseError(code, path, section, error_index, error_pos, error_msg) => return ErrorBuilder.new()
+				.render(code, path, section, error_index, error_pos, error_msg)
+				.exit(3);
 		}
+		modules[i] = module;
+		names[i] = extractName(path);
+	}
+	var err = ErrorGen.new(all_paths[all_paths.length - 1]);
+	// Call monitors for all modules
+	for (i < modules.length) {
```
titzer, line 82: I like using the `for (x in array)` syntax whereever possible, and it looks like it will work here.

test/linker/test.sh
```diff
@@ -0,0 +1,65 @@
...
+
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do
+    DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
+    SOURCE="$(readlink "$SOURCE")"
+    [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE"
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
+
+. $DIR/../common.sh R3-monitor
```
titzer, line 11: Should be "linker" here

titzer (commented):
First round of comments. There are some formatting issues and I'll want to take a look deeper into the Instantiator changes.

test/linker/test.sh
thelongmarch-azx: yet another example of "software engineering is just copy & pasting"

src/wizeng.main.v3
```diff
@@ -283,6 +284,7 @@ component WizengOptions {
 	def group = OptionsRegistry.newGroup("WIZENG", null);
 	def HELP = group.newFlag("help", printHelp, "Print command-line help.");
 	def HELP2 = group.newFlag("-help", printHelp, "Print command-line help.");
+	def PROG_ARGS = group.newOption("-args", null, "=<arg_1, ... ,arg_n>", parseArgs, "Provide a list of program args for certain host module initialization.");
```
thelongmarch-azx, line 287: programArgs used to be optional args appended after the single wasm file is provided. Now that we are accepting multiple wasm files, I made programArgs an explicit option to avoid mistaking it for wasm files

titzer: I think we should keep the behavior that arguments after the `.wasm` files are arguments to the program; this is standard practice. If there is ambiguity between the end of the list of `.wasm` files and the start of the arguments to the (main) module, then most engines (including JS engines) use `--` to slit the "program" from its arguments. Let's keep the program args behavior for now and implement the `--` behavior in the next PR.

titzer (approved):
I think this is good to merge. We can always iterate with minor cleanups.

PR #208, opened 2024-10-11
WasmParser duplicator

matthew-mojira:
This is a utility for the WhammMonitor, which needs to clone the current position of a CodeValidator as it parses through a wasm module

### DIFF
--- a/src/engine/WasmParser.v3
+++ b/src/engine/WasmParser.v3
@@ -505,6 +505,12 @@ class WasmParser(extensions: Extension.set, limits: Limits, module: Module,
 		if (b < ' ' || b > 127) b = '.';
 		return b;
 	}
+
+	def dup() -> WasmParser {
+	  // CodePtr should be able to dup itself?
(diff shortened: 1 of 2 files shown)


src/engine/WasmParser.v3
```diff
@@ -505,9 +505,15 @@ class WasmParser(extensions: Extension.set, limits: Limits, module: Module,
 		if (b < ' ' || b > 127) b = '.';
 		return b;
 	}
+
+	def dup() -> WasmParser {
```
titzer, line 509: Comment sounds good to me; want to roll a `CodePtr.dup()` into this PR?

src/engine/CodePtr.v3
```diff
@@ -218,4 +218,7 @@ class CodePtr extends DataReader {
 		skip_immediates(opcode.imms);
 		return opcode;
 	}
+	def dup() -> CodePtr {
+		return CodePtr.new(null).reset(Ranges.dup(data), pos, limit);
```
titzer, line 222: No need to duplicate the data, just the `CodePtr`. Other than that, looks good.

titzer (approved):

PR #209, opened 2024-10-11
Refractored suspension handler out of the exception handler lookup

linxuanm:
This PR refractors suspension handler lookup out of the exception handler lookup routine. I think this makes more sense as this limits the search for a handler to a more limited scope rather than the whole function declaration, and is in general cleaner as the lookup of exception handler and suspension handler aren't really that similar.

### DIFF
--- a/src/engine/BytecodeIterator.v3
+++ b/src/engine/BytecodeIterator.v3
@@ -745,9 +745,9 @@ class BytecodeIterator {
 
 			CONT_NEW => 			v.visit_CONT_NEW(read_CONT());
 			CONT_BIND => 			v.visit_CONT_BIND(read_CONT(), read_CONT());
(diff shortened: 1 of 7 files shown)


titzer (commented):
I'm not convinced that suspension is sufficiently different from exception handling to warrant a completely different mechanism. This seems like it could be a lot of added complexity, and using the side table is very low level and very buggy.

With this change, it no longer handles the target stack delta (i.e. popcount) in sidetable entries properly anymore.

I think we should hold off on refactoring the mechanism until all the stack switching tests pass and they are turned on as part of the regression testing. Otherwise we're going to introduce a bunch of bugs that we have to chase down later.

titzer:
Going to close this now, as I think this is covered in the current stack switching changes.

PR #210, opened 2024-10-20
[stack-switching] Additional spectests

linxuanm:
This PR adds some spectests for stack switching as per the TODO and some other edge cases.


titzer (approved):
lgtm, rubber-stamped

PR #211, opened 2024-10-21
[stack-switching] Heaptypes in continuation params + duplicate handler

linxuanm:
This PR adds more spectests for stack switching as per the TODO.

In `resume17.wast`, there are two handler entries (which points to the same jump position for now, but they can be made different) and this seems to be allowed (I don't recall any checks on duplicate exception/suspension handlers in Wizard's implementation, and WasmFX permits this and uses the first handler). I'm putting this in to verify that this is indeed the right behavior, although I'm not sure if this belongs in a spectest. Please let me know if it should be removed.

TODO (for me): add more tests for other GC heaptypes


titzer:
> This PR adds more spectests for stack switching as per the TODO.
> 
> In `resume17.wast`, there are two handler entries (which points to the same jump position for now, but they can be made different) and this seems to be allowed (I don't recall any checks on duplicate exception/suspension handlers in Wizard's implementation, and WasmFX permits this and uses the first handler). I'm putting this in to verify that this is indeed the right behavior, although I'm not sure if this belongs in a spectest. Please let me know if it should be removed.
> 

No, sounds like a good thing to test, especially if it passes the reference interpreter and wasmtime.

linxuanm:
This should be all for this PR for now. I'm thinking of also adding tests for throwing & catching on different stacks (maybe in a future PR? I know for sure that this doesn't work on fast interpreter for now).

titzer:
> This should be all for this PR for now. I'm thinking of also adding tests for throwing & catching on different stacks (maybe in a future PR? I know for sure that this doesn't work on fast interpreter for now).

Those sound like good tests too. I'll merge this one now though.

PR #213, opened 2024-10-23
[stack-switching] Nested resume with suspensions handled at different levels

linxuanm:
A while ago I made a terrible mistake when reconstructing the links between stacks during a resume, which led to an hour of debug today. These tests are here to specifically address that.


titzer (approved):
lgtm; rubber-stamped

PR #215, opened 2024-10-26
[stack-switching] Updated bytecode reading to accept switch handling

linxuanm:
An update in the stack switching proposal changed the bytecode format for `resume` and `resume_throw`. This PR updates the corresponding parts in Wizard.

### DIFF
--- a/src/engine/CodePtr.v3
+++ b/src/engine/CodePtr.v3
@@ -184,11 +184,19 @@ class CodePtr extends DataReader {
 					var count = read_uleb32();
 					for (i < count + 1) skip_leb();
 				}
-				HANDLERS => {
+				EX_HANDLERS => {
 					var count = read_uleb32();
 					for (i < count) {
 						skip_leb();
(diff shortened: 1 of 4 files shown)


titzer (commented):
Let's add some tests for this in `test/regress/ext:stack-switching` and check in the `.bin.wast` for them.

linxuanm:
Actually after checking it seems that the spectests were already updated (compiled with new version ref interpreter). However since most of the tests that involve resume/suspend features only one handler that has a single entry, which handles tag `#0` and jumps to the inner-most block of depth `0`. This resulted in a lot of tests being passed even when the new code validator isn't implemented (as a lot of the handlers have sequence `\00\00\00`).

titzer:
Should I merge this PR, or is its logic included in the #216 ?

linxuanm:
#216 is rebased onto this branch. I think we can work on #216 and ignore this for now. Sorry for the confusion.

PR #216, opened 2024-10-26
Fixing V3 interpreter for regression tests

linxuanm:
This PR aims to make the V3 interpreter pass the regression tests we have so far


src/engine/Debug.v3
```diff
@@ -4,7 +4,7 @@
 component Debug {
 	def paranoid = false;
 	def verbose = false;
-	def interpreter = false;
+	def interpreter = true;
```
titzer, line 7: This is left over.

linxuanm:
TODO: check implicit popping after a `ctlxfer` with `resume`.

linxuanm:
This PR is done; however, the checks do not pass as spectests do not pass for x86_64 target yet. Should we disable them for the time being (I don't think this can be done in `regress.sh` easily).

titzer:
> This PR is done; however, the checks do not pass as spectests do not pass for x86_64 target yet. Should we disable them for the time being (I don't think this can be done in `regress.sh` easily).

Yes, it's possible to add them to `test/failures.x86-64-linux`.

PR #218, opened 2024-11-04
Updated Whamm monitor

matthew-mojira:


src/monitors/WhammMonitor.v3
```diff
@@ -66,46 +70,226 @@ class WhammMonitor(whamm: Module) extends Monitor {
...
-				for (l = bp; l != null; l = l.tail) {
-					if (Trace.whamm) Trace.OUT.put3("  match func %q+%d: %s", func.render(module.names, _), bi.pc, bi.current().mnemonic).ln();
-					var wf = WasmFunction.!(wi.functions[bp.head.2.func_index]);
-					var args = bindParams(it, bp.head.1);
-					Instrumentation.insertLocalProbe(module, func.func_index, bi.pc, WhammProbe.new(wf, args));
-				}
-			}
+
+      // parsed module is already validated, so we don't expect a new error here
+	    var cv = CodeValidator.new(Extension.WIZENG, Limits.new(), module, ErrorGen.new(""));
```
titzer, line 90: I think we should pass in the outer error gen; even if we expect nothing to go wrong, it's probably better to not swallow any errors here.

src/monitors/WhammMonitor.v3
```diff
@@ -66,46 +70,226 @@ class WhammMonitor(whamm: Module) extends Monitor {
...
-				var bp = opcodePatterns[bi.current().tag];
-				for (l = bp; l != null; l = l.tail) {
-					if (Trace.whamm) Trace.OUT.put3("  match func %q+%d: %s", func.render(module.names, _), bi.pc, bi.current().mnemonic).ln();
-					var wf = WasmFunction.!(wi.functions[bp.head.2.func_index]);
-					var args = bindParams(it, bp.head.1);
-					Instrumentation.insertLocalProbe(module, func.func_index, bi.pc, WhammProbe.new(wf, args));
-				}
-			}
+
+      // parsed module is already validated, so we don't expect a new error here
```
titzer, line 89: Something is off in the indentation here. Please check you are using tabs and the tab stops are set at 8.

src/monitors/WhammMonitor.v3
```diff
@@ -66,46 +70,226 @@ class WhammMonitor(whamm: Module) extends Monitor {
...
+				Arg(orig, i)   => {  // arg0 = first arg to bytecode
+				  def io = Opcodes.stackDelta(cv.sig, op, dupParser(cv));
+				  def args = io.0;
+				  if (i >= args) {
+				    match (op) {
+              CALL, CALL_INDIRECT => return null; // no match
+              _ => err.at(cv.parser.decoder).WhammProbeError("whamm probe", "instruction argument out of bounds");
+				    }
+				  }
+			    def stk = cv.val_stack.elems;
```
titzer, line 221: I think this will be a little more robust to do a `.peekn` and use the parameter arity from `stackDelta`.

src/monitors/WhammMonitor.v3
```diff
@@ -66,46 +70,226 @@ class WhammMonitor(whamm: Module) extends Monitor {
...
+			cv.validateFunc(func);
+		}
+	}
+
+	def bytecodeProbe(module: Module, wi: Instance, func: FuncDecl, cv: CodeValidator, opcode_pos: int, opcode: Opcode) {
+    var bp = opcodePatterns[opcode.tag];
+
+		for (l = bp; l != null; l = l.tail) { // for every probe
+  		if (Trace.whamm) Trace.OUT.put3("  match func %q+%d: %s", func.render(module.names, _), opcode_pos, opcode.mnemonic).ln();
+			var wf   = WasmFunction.!(wi.functions[bp.head.2.func_index]);
```
titzer, line 101: Could be an imported function as well. I think we only need the signature of the function, correct?

matthew-mojira: I am not sure what you mean by this

titzer: The cast to `WasmFunction.!` is overly restrictive here; you just need the function's signature IIUC.

titzer (approved):
lgtm

PR #219, opened 2024-11-08
[gc] Fix a few bugs in parsing and validation

vouillon:


titzer (commented):
Thanks!

Can you add a test in `test/regress/ext:gc`?

titzer:
Thanks for this PR. Because it needed some followup work, I've fixed the unit tests and added regression tests for this in #221 .

PR #222, opened 2024-11-12
Whamm intrinsification baseline (wasm to wasm call)

thelongmarch-azx:


src/engine/compiler/SinglePassCompiler.v3
```diff
@@ -1457,6 +1536,9 @@ class SinglePassCompiler(xenv: SpcExecEnv, masm: MacroAssembler, regAlloc: RegAl
 		var pc = it.pc - start_pos;
 		instrTracer.instr_width = Opcodes.longestName + 1;
 		instrTracer.putPcAndInstr(OUT, module, func, pc, orig);
+		OUT.puts("JIT code: ");
+		masm.printCodeBytes(OUT);
+		OUT.ln();
```
thelongmarch-azx, line 1541: I think this makes debugging easier, since we can see the exact assembly generated for each opcode

src/engine/x86-64/X86_64Runtime.v3
```diff
@@ -106,6 +106,11 @@ component X86_64Runtime {
 		if (ret != null) return stack.throw(ret);
 		return ret;
 	}
+	def runtime_materialize_frame_accessor() {
```
titzer, line 109: How about calling this `runtime_getFrameAccessorMetaRef()`and returning the metaref, since it's needed by the caller?

src/engine/x86-64/X86_64MacroAssembler.v3
```diff
@@ -21,11 +21,19 @@ class X86_64MacroAssembler extends MacroAssembler {
 	var jump_tables: Vector<(int, Array<X86_64Label>)>;
 	var offsets: V3Offsets;
 	var trap_stubs: X86_64SpcTrapsStub;
+	var last_code_print_at = 0;
 
 	new(w, regConfig: RegConfig) super(Target.tagging, regConfig) {
 		scratch = G(regConfig.scratch);
 	}
-
+	def printCodeBytes(sb: StringBuilder) {
```
titzer, line 29: This adds a little more local state; how about keeping track of the offset of the bytecode start in the SPC? Partially because we should consider saving that for use at runtime, e.g. to compute the PC at frame walking instead of storing it as is done now.

src/engine/x86-64/X86_64Target.v3
```diff
@@ -39,6 +39,13 @@ component Target {
 		if (Trace.compiler) {
 			Trace.OUT.put2("func[%d].target_code: break *0x%x", f.func_index, addr - Pointer.NULL)
 				.put2(" disass 0x%x, 0x%x", addr - Pointer.NULL, end - Pointer.NULL).ln();
+			var cur_byte = addr;
```
titzer, line 42: How about guarding this behind a new trace flag `--trace-asm`?

titzer (approved):
Looks good with minor suggestions.

titzer (approved):

PR #228, opened 2024-11-21
Remove excessive spaces for parsing opcode mnemonics

matthew-mojira:
Updated `parseName` to handle `ref.test null` and `ref.cast null` without excessively eating spaces for opcodes which do not need it. I need it to parse opcodes Whamm probes as there can be spaces between the opcode mnemonic and the predicate separator.

### DIFF
--- a/src/engine/Opcodes.v3
+++ b/src/engine/Opcodes.v3
@@ -799,10 +799,27 @@ component Opcodes {
 			if (ch >= '0' && ch <= '9') continue;
 			if (ch == '_') continue;
 			if (ch == '.') continue;
-			if (ch == ' ') continue; // TODO: handles ref.test null
 			break;
 		}
-		if (p > 0) return findByName(Ranges.dup(str[0 ... p]));
+		if (p > 0) {
+			def op = findByName(Ranges.dup(str[0 ... p]));
+			// ref.test and ref.cast have null versions with a space in the opcode name,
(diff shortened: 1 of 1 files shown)


src/engine/Opcodes.v3
```diff
@@ -799,10 +799,27 @@ component Opcodes {
...
 			break;
 		}
-		if (p > 0) return findByName(Ranges.dup(str[0 ... p]));
+		if (p > 0) {
+			def op = findByName(Ranges.dup(str[0 ... p]));
+			// ref.test and ref.cast have null versions with a space in the opcode name,
+			// so we check those opcodes to see if it has null at the end
+			match (op) {
+				REF_TEST => {
+					if (str.length >= p + 4 && Ranges.equal(str[p ... (p + 5)], " null")) {
```
titzer, line 810: You can simplify this to `Ranges.equal(str[p ...], " null")` and combine the two cases with a `||`

titzer (approved):
Looks good with minor simplification.

src/engine/Opcodes.v3
matthew-mojira: I don't believe this keeps the same behavior. I would like to see if " null" prefixes `str[p ...]`.

titzer: Sure, I see what you are saying. Let's just land this and can polish it a bit later.

PR #226, opened 2024-11-20
Static predicates in Whamm probes

matthew-mojira:

PR #230, opened 2024-11-22
Fix failing multi-memory test

ejrgilbert:


ejrgilbert:
This test should work after PR #226 is merged.

ejrgilbert:
@titzer What would happen if you're needing to use `puts` in a multi-memory context?

matthew-mojira:
What a great PR!

titzer:
> @titzer What would happen if you're needing to use `puts` in a multi-memory context?

`wizeng.puts` tries to bind to the first exported memory of a module. The logic is in `src/modules/wizeng/WizengModule.v3`.

PR #239, opened 2024-12-11
write whamm hotness-mon.wasm

ejrgilbert:
I created a generator for this that we can check in so that as opcodes are added, we can just rerun the generator and it'll pick up those new opcodes.

If you don't want it in the repo, I can just make a new repo to keep track of items like this.


test/monitors/whamm/monitors/generators/gen_hotness-mon/GenHotnessMon.v3
```diff
@@ -0,0 +1,247 @@
...
+    writet(fd, 2, "i64.add\n");
+    writet(fd, 2, "i64.store offset=12\n");
+
+    func_end(fd);
+}
+
+def flush_func(fd: int) {
+    func_start(fd, "$flush (export \"wasm:exit\")", true);
+    
+    writet(fd, 2, "(local $entry i32)\n");
```
titzer, line 126: There are lots of options for making this nicer, e.g. using a partially-applied function:

```
  var w = writet(fd, 2, _);
  w("line 1");
  w("line 2");
  w("line 3");
  w("line 4");
  w = writet(fd, 3, _);
  w("line 5");
  w("line 6");
  w("line 7");
```

You can make this a little nicer with a helper class that has the indentation state in it and use the "return this" pattern, like so:

```
var o = IndentingOutputter.new(); // choose your favorite name here

o.indent()
  .w("level 1 stuff")
  .w("level 1 stuff")
  .w("level 1 stuff")
  .indent()
    .w("level 2 stuff")
    .unindent()
```

Or other options are to take an array of strings and you can pass a literal:

```
 o.w([
  "line 1",
  "line 1",
  "line 1"
  ])

I usually just do this to get rid of as much noise as possible. Since the generator's performance doesn't matter at all, you can make it nice.

```

titzer (approved):
lgtm modulo some Virgil-ization.

PR #242, opened 2024-12-12
[stack-switching] `resume_throw` Implementation

linxuanm:
This PR implements `resume_throw`. It also cleans up the previous stack switching code + stack stubs in preparation for JIT implementation.


titzer (approved):
I'm going to land this as a mostly rubber-stamped in order to base some refactorings on top of it. In particular, I am going to remove the abrupt return mechanism and rely completely on stack unwinding--it's the default now and exception handling doesn't work without it.

PR #245, opened 2024-12-17
Enable linking for whamm monitors

ejrgilbert:
Still having an issue with the `linking_static-pred.wat` test due to `puts` printing from the whamm monitor's memory (it's outputting "hello call!" twice instead of printing from the linked `base-instr.wat` module's memory, which should print "hello pred!"). Maybe we should just change this test to not have a print in the predicate and just verify that the linking behavior is correct with a print in the main module instead?

I can make another PR to do this ^^


src/monitors/MonitorOptions.v3
```diff
@@ -44,21 +44,59 @@ component MonitorOptions {
...
-			match (result) {
-				Ok(m) => {
-					monitors.put(WhammMonitor.new(m));
-					return;
+
+			// Split by '+'
+			var s = 0, pos = s;
+			var whammModuleName = "";
+			var otherModuleNames = Vector<string>.new();
+			while (pos < name.length) {
```
titzer, line 52: I think this is a good candidate for adding to the Virgil `Strings` utility: `Strings.split(string, byte) -> Array<string>`. Wdyt?

src/monitors/MonitorOptions.v3
```diff
@@ -44,21 +44,59 @@ component MonitorOptions {
...
-					.ln()
-					.exit<void>();
-				ParseError(code, path, section, error_index, error_pos, error_msg) => return ErrorBuilder.new()
-					.render(code, path, section, error_index, error_pos, error_msg)
-					.exit<void>();
 			}
+			// If whamm module hasn't been set yet, we don't have any libraries to link
+			var finalName = Arrays.range(name, s, pos);
+			if (!Strings.endsWith(finalName, ".wasm")) {
+				ErrorBuilder.new()
```
titzer, line 72: I feel like this is a good time and bite the bullet to feed in an `ErrorGen` to `parse()` above and feed it here. Then it's the caller's responsibility to pass one in and report errors.

src/monitors/WhammMonitor.v3
```diff
@@ -358,13 +358,47 @@ class WhammMonitor(whamm: Module) extends Monitor {
...
-		var imports = Array<Exportable>.new(whamm.imports.length), err = ErrorGen.new(whamm.filename);
-		var i = Instantiator.new(Extension.set.all, whamm, imports, err);
-		i.processors.put(WizengModule.new());
+		if (Trace.whamm) Trace.OUT.put2("Instantiating whamm module %s for %s", whamm.1.filename, module.filename).ln();
+		var imports = Array<Exportable>.new(whamm.1.imports.length), err = ErrorGen.new(whamm.1.filename);
+
+		// Perform the linking for Whamm monitor support libraries.
+		var engine = Engine.new();
+		engine.extensions = Extension.set.all;
+		if (engine.extensions.WIZENG) ModuleRegistry.expose("wizeng");
```
titzer, line 367: Is it possible to add the `wizeng` module to the linking context?

src/util/PathUtil.v3
```diff
@@ -0,0 +1,15 @@
+component PathUtil {
```
titzer, line 1: This is now a little overfit to the extension; it trims off the last 5 characters. To generalize this it should probably scan forward for the last slash and then scan backward for the last '.'. We can also do that without making a string buffer, just keeping track of the start and end and using either `Arrays.range()` or `Ranges.dup(path[start ... end])`.

src/monitors/MonitorOptions.v3
ejrgilbert: https://github.com/titzer/virgil/pull/326

Once this is merged, I'll address the rest of the comments here!

src/modules/ModuleLinker.v3
```diff
@@ -16,6 +16,9 @@ class ModuleLinker(engine: Engine,
 	private def deferred_imports = Array<Array<int>>.new(named_modules.length);
 
     def link() -> Array<Instantiator> {
+		// Expose the wizeng HostModule if it's an activated extension
+		if (engine.extensions.WIZENG) ModuleRegistry.expose("wizeng");
```
ejrgilbert, line 20: I'm not sure what you meant by adding the wizeng HostModule to the linking context [here](https://github.com/titzer/wizard-engine/pull/245#discussion_r1890905079), I pushed it down into the linker.link() logic in case this is what you meant.

titzer (approved):
lgtm, nice!

PR #250, opened 2025-01-13
[wasi] Fix statx syscall, bad 'mask' arg

ejrgilbert:


src/modules/wasi/WasiOptions.v3
```diff
@@ -54,4 +55,17 @@ component WasiOptions {
 	def getEnvs() -> Vector<string> {
 		return envs;
 	}
+	def printHelp(out: TraceBuilder) {
```
titzer, line 58: Nice, thanks for adding help for these options!

PR #280, opened 2025-02-08
[stack-switching] Fix incorrect exception handling after `resume` (#273)

### LINKED ISSUE #273: Bad interaction between stack-switching and exception handling
The following test fails. (It pass with the reference interpreter.)
```wasm
(module
  (tag $exn)
  (type $fun (func))
  (type $cont (cont $fun))
  (elem declare func $g)
  (func (export "main")
    (block $exit
      (try_table (catch $exn $exit)
        (resume $cont (cont.new $cont (ref.func $g))))
      (throw $exn)))
  (func $g)
)

(assert_exception (invoke "main"))
```
Here is the binary version:
```wasm
(module definition binary
  "\00\61\73\6d\01\00\00\00\01\86\80\80\80\00\02\60"
  "\00\00\5d\00\03\83\80\80\80\00\02\00\00\0d\83\80"
  "\80\80\00\01\00\00\07\88\80\80\80\00\01\04\6d\61"
  "\69\6e\00\00\09\85\80\80\80\00\01\03\00\01\01\0a"
  "\a2\80\80\80\00\02\95\80\80\80\00\00\02\40\1f\40"
  "\01\00\00\00\d2\01\e0\01\e3\01\00\0b\08\00\0b\0b"
  "\82\80\80\80\00\00\0b"
)
(module instance)
(assert_exception (invoke "main"))
```
I get the following output:
```
$ ./bin/spectest.x86-linux -ext:stack-switching ./bug.bin.wast 
##+./bug.bin.wast
##-fail: ./bug.bin.wast @ 11:2 assert_exception expected exception, got ()
```
Somehow, the exception gets caught by the exception handler, even thought the `throw` is not in scope. I suspect `prev_resume_pc` is not reset when an continuation terminates normally (rather than suspending itself).

linxuanm:
This PR fixes #273, which incorrectly catches an exception outside the range of a handler.


titzer (approved):
lgtm, thanks!

PR #282, opened 2025-02-10
WASI: fix path_open

vouillon:
The Unix file descriptor was returned instead of the user-level file descriptor.


titzer (approved):
nice catch!

PR #289, opened 2025-02-10
[fast-int] Fix large blocktype LEB

titzer:
Code to skip extended block types was not handling LEBs properly.


titzer:
Addresses #289

PR #290, opened 2025-02-11
[stack-switching] `switch` support in common code

linxuanm:
This PR adds support for switch in code common to V3Int, fast-int and JIT.

### DIFF
--- a/src/engine/Execute.v3
+++ b/src/engine/Execute.v3
@@ -25,7 +25,7 @@ component Execute {
 }
 
 type StackSwitchInfo {
-	case Suspend(top: WasmStack, tagdef: Tag, vals: Array<Value>) {}
+	case Suspend(top: WasmStack, tagdef: Tag, vals: Array<Value>, is_switch: bool) {}
 	case RunChildStack(cont: Continuation) {}
 }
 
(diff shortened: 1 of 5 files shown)


src/engine/Execute.v3
```diff
@@ -25,7 +25,7 @@ component Execute {
 }
 
 type StackSwitchInfo {
-	case Suspend(top: WasmStack, tagdef: Tag, vals: Array<Value>) {}
+	case Suspend(top: WasmStack, tagdef: Tag, vals: Array<Value>, is_switch: bool) {}
```
titzer, line 28: Hmm, it might be cleaner to introduce a new ADT case here, rather than a bool switch. Wdyt?

src/engine/v3/V3Interpreter.v3
```diff
@@ -1493,7 +1493,10 @@ class V3Interpreter extends WasmStack {
 		}
 	}
 	// Returns whether the suspend got handled.
-	def tryHandleSuspend(cont: Continuation, tag: Tag, vals: Array<Value>) -> bool {
+	def tryHandleSuspend(cont: Continuation, tag: Tag, vals: Array<Value>, is_switch: bool) -> bool {
```
titzer, line 1496: If we add a new case, then we could also split this to a different function and factor out some commonalities between the two. I think that's cleaner than a `bool` that changes behavior.

src/engine/Execute.v3
linxuanm: I think this is a good idea. I just finished the V3Int implementation of `switch` and currently I'm abusing the stack to pass continuation values around. With a new `switch` case this would be much cleaner.

PR #295, opened 2025-02-13
[wasi] Implement more functions

vouillon:
Also, fix timestamps in the filestat record (depend on titzer/virgil#352) and update path_open to support more flags.

### DIFF
--- a/src/modules/wasi/x86-64-linux/WspOneModule-x86-64-linux.v3
+++ b/src/modules/wasi/x86-64-linux/WspOneModule-x86-64-linux.v3
@@ -5,6 +5,13 @@ def OUT = Trace.OUT;
 def X_ = WspOneRegistration.register(X86_64Linux_WspOneModule.new());
 // Linux system call numbers
 def SYS_openat = 257;
+def SYS_mkdirat = 258;
+def SYS_unlinkat = 263;
+def SYS_renameat = 264;
+def SYS_linkat = 265;
+def SYS_symlinkat = 266;
+def SYS_readlinkat = 267;
+def SYS_utimensat = 280;
 def SYS_clock_get_time = 228;
 def SYS_clock_get_res = 229;
 def SYS_getrandom = 318;
@@ -15,9 +22,13 @@ def SYS_sched_yield = 24;
 def SYS_kill = 37;
 def SYS_getpid = 39;
 def SYS_statx = 332;
+def SYS_ftruncate = 77;
+def SYS_getdents = 78;
 
 def STATX_ALL = 0x00000FFF; // all currently-supported flags
 def AT_SYMLINK_NOFOLLOW = 0x100;
+def AT_SYMLINK_FOLLOW = 0x400;
+def AT_EMPTY_PATH = 0x1000;
 
 // Linux lseek whence values
 def WHENCE_SET = 0;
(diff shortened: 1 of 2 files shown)


src/modules/wasi/x86-64-linux/WspOneModule-x86-64-linux.v3
```diff
@@ -147,6 +163,136 @@ class X86_64Linux_WspOneModule extends WspOneModule {
...
 		return HostResult.Value1(Values.I32_0);
 	}
+        def path_unlink_file(args: Range<Value>) -> HostResult {
+		// 0=fd 1=pathptr 2=length
+		var fd		= Values.unbox_i(args[0]);
+		var pathptr	= args[1];
+		var pathlen	= args[2];
+
+		var dir_fd = fdmap.get(fd);
+		if (dir_fd == null) return HostResult.Value1(Values.box_i(WspOneErrno.BADF.code));
```
titzer, line 173: This line (which is a copied a few times from a previous example) can use the `i` helper, or even a constant factored out to the top-level.

src/modules/wasi/x86-64-linux/WspOneModule-x86-64-linux.v3
```diff
@@ -169,11 +315,19 @@ class X86_64Linux_WspOneModule extends WspOneModule {
...
 
-		if ((oflags & (1u << WspOneFdFlags.creat.tag)) != 0) sysflags |= LinuxConst.O_CREAT;
-		if ((oflags & (1u << WspOneFdFlags.trunc.tag)) != 0) sysflags |= LinuxConst.O_TRUNC;
-		if ((oflags & (1u << WspOneFdFlags.excl.tag)) != 0) sysflags |= LinuxConst.O_EXCL;
+		if ((oflags & (1u << WspOneOFlags.creat.tag)) != 0) sysflags |= LinuxConst.O_CREAT;
+		if ((oflags & (1u << WspOneOFlags.directory.tag)) != 0) sysflags |= LinuxConst.O_DIRECTORY;
+		if ((oflags & (1u << WspOneOFlags.excl.tag)) != 0) sysflags |= LinuxConst.O_EXCL;
+		if ((oflags & (1u << WspOneOFlags.trunc.tag)) != 0) sysflags |= LinuxConst.O_TRUNC;
+		if ((fflags & (1u << WspOneFdFlags.append.tag)) != 0) sysflags |= LinuxConst.O_APPEND;
+//		if ((fflags & (1u << WspOneFdFlags.dsync.tag)) != 0) sysflags |= LinuxConst.O_DSYNC;
```
titzer, line 327: Are these TODOs are unused? If they are unused, we should drop them.

titzer (commented):
Thanks for doing all this work!

Do you have any wasm modules that could serve as test cases? There is the wasi testsuite, which there are scripts to pull and run. Do any of those tests now pass?

I'm happy to accept this PR as correct, but want to avoid regressing as we refactor to reduce the verbosity of the unboxing and checking with more utilities.

titzer:
Merging this and will follow up because dependent Virgil changes break build otherwise.

vouillon:
At the moment, [most of the Js_of_ocaml / Wasm_of_ocaml test suite is passing](https://github.com/ocsigen/js_of_ocaml/actions/runs/13333694834/job/37243854692), including an implementation of OCaml effects handlers based on the stack switching proposal, which is pretty cool. It is quite large but just a few tests are failing.
But it's not really easy to extract from this some wasm modules that could serve as test cases...

src/modules/wasi/x86-64-linux/WspOneModule-x86-64-linux.v3
vouillon: These are more like TODOs.

titzer:
> At the moment, [most of the Js_of_ocaml / Wasm_of_ocaml test suite is passing](https://github.com/ocsigen/js_of_ocaml/actions/runs/13333694834/job/37243854692), including an implementation of OCaml effects handlers based on the stack switching proposal, which is pretty cool. It is quite large but just a few tests are failing. But it's not really easy to extract from this some wasm modules that could serve as test cases...

Wow, that's awesome! I am happy to add it as another optional suite of tests in the same way that the Virgil test suite can be separately compiled to wasm and run (See https://github.com/titzer/wizard-engine/tree/master/test/virgil).

As it happens, I have some students working on test case extraction from larger programs, so they might be able to apply it to these tests.

PR #300, opened 2025-02-14
[switch-switch] V3 refactor

linxuanm:
This PR simplifies the stack return states.


titzer (approved):
Very nice simplification!

PR #301, opened 2025-02-14
[stack-switching] Fixed `switch` not consuming continuation

linxuanm:
Fixed a silly bug. Previous version of `switch` didn't consume the continuation.


titzer (approved):
Thanks. Can you also write a quick test for this?

titzer (approved):

titzer:
I turned on `spc` mode for x86-64-linux, so you need to add the new test to `test/regress.failures.x86-64-linux.spc`.

linxuanm:
Thanks for the heads up!

PR #306, opened 2025-02-17
Bugfix/wasi state

ejrgilbert:
I'm not sure how to keep from duplicating the fdMap logic between the WspOneModule and the WspOneModule-<platform> files. Do you have an idea?


src/modules/wasi/x86-64-linux/WspOneModule-x86-64-linux.v3
```diff
@@ -2,7 +2,70 @@
 // See LICENSE for details of Apache 2.0 license.
 
 def OUT = Trace.OUT;
-def X_ = WspOneRegistration.register(X86_64Linux_WspOneModule.new());
+def X_ = WspOneRegistration.add(registerX86_64Linux_WspOneModule);
```
titzer, line 5: I think we can fix this duplication between the target-independent module and this target-dependent module in a similar way to how the `Target` component is defined in each of the targets. For example, we could have a `WasiTarget` component that is defined one way for the V3 targets and another way for the `x86-64-linux` target. That component could have method to allocate an instance of the `WasiInstance`.

PR #316, opened 2025-02-22
[stack-switching] fast-int `switch` implementation

linxuanm:
This PR implementsn `switch` in the fast-interpreter.

TODO: I think it would be best to refactor the runtime function for `suspend` and `switch` into x86 and leave a common runtime function for unwinding the stack chain until the tag is handled (which requires a runtime call to `X86_64Stack` and isn't doable in x86 along).


titzer (commented):
Can you update the expected failure files?

linxuanm:
I created a new PR with more progress on the proposed refactor changes I mentioned above. Let's work with that one.

PR #321, opened 2025-02-23
[stack-switching] `switch` implementation

linxuanm:
This PR implements `switch` in the fast-interpreter.

### DIFF
--- a/src/engine/Module.v3
+++ b/src/engine/Module.v3
@@ -170,11 +170,11 @@ class FuncDecl(sig_index: int) extends Decl {
 	def findExHandler(instance: Instance, tag: Tag, throw_pc: int) -> ExHandler {
 		return findHandler("findExHandler", handlers, instance, tag, throw_pc);
 	}
-	def findSuspensionHandler(kind: SuspensionType, instance: Instance, tag: Tag, resume_pc: int) -> ExHandler {
-		match (kind) {
-			SUSPEND => return findHandler("suspend", suspend_handlers, instance, tag, resume_pc);
-			SWITCH => return findHandler("switch", switch_handlers, instance, tag, resume_pc);
-		}
+	def findSuspensionHandler(instance: Instance, tag: Tag, resume_pc: int) -> ExHandler {
+		return findHandler("suspend", suspend_handlers, instance, tag, resume_pc);
+	}
+	def findSwitchHandler(instance: Instance, tag: Tag, resume_pc: int) -> ExHandler {
+		return findHandler("switch", switch_handlers, instance, tag, resume_pc);
 	}
 	def findHandler(name: string, handlers: Array<ExHandlerEntry>, instance: Instance, tag: Tag, throw_pc: int) -> ExHandler {
 		var i = 0;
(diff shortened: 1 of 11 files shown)


src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -3249,6 +3273,11 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
 			asm.movq_r_m(r_dispatch, masm.absPointer(offsets.Interpreter_dispatchTable));
 		}
 	}
+	// called after switching to a new stack
```
titzer, line 3276: I think I prefer this to remain inline; otherwise it seems to imply there is something special about stack switches, which doesn't seem to be the case.

src/engine/Module.v3
```diff
@@ -170,11 +170,11 @@ class FuncDecl(sig_index: int) extends Decl {
 	def findExHandler(instance: Instance, tag: Tag, throw_pc: int) -> ExHandler {
 		return findHandler("findExHandler", handlers, instance, tag, throw_pc);
 	}
-	def findSuspendHandler(instance: Instance, tag: Tag, resume_pc: int) -> ExHandler {
-		return findHandler("SusHandler", suspend_handlers, instance, tag, resume_pc);
-	}
-	def findSwitchHandler(instance: Instance, tag: Tag, resume_pc: int) -> ExHandler {
-		return findHandler("switch", switch_handlers, instance, tag, resume_pc);
+	def findSuspensionHandler(kind: SuspensionType, instance: Instance, tag: Tag, resume_pc: int) -> ExHandler {
```
titzer, line 173: AFAICT the suspension type is only used here to select a string for tracing. If we have tracing further up in callers (e.g. a line of output to indicate whether a suspend or a switch is happening), then we can simplify this a bit.

src/engine/Runtime.v3
```diff
@@ -350,7 +350,19 @@ component Runtime {
 		stack.push(Value.Ref(new_cont));
 		return null;
 	}
-
+	// Finds a handler for {tag} by unwinding the stack-switching chain and return
+	// a freshly created continuation with {Continuation.bottom} pointing to the
+	// direct child of the handler stack.
+	def unwindStackChain(stack: WasmStack, instance: Instance, tag: u31, unwind_type: SuspensionType) -> Continuation {
```
titzer, line 356: Similarly here, if `findSuspensionHandler` doesn't take the suspension type anymore, then this method won't need it.

src/engine/x86-64/X86_64Stack.v3
```diff
@@ -151,32 +152,39 @@ class X86_64Stack extends WasmStack {
 	def setRsp(ptr: Pointer) -> this {
 		rsp = ptr;
 	}
-	// This function returns true if a handler IS NOT found, and otherwise
-	// returns {(stp_idx, order)}.
-	def findSuspendHandler(frame: TargetFrame, tag: Tag) -> bool {
+	def findSuspensionHandler(kind: SuspensionType, tag: Tag) -> ExHandler {
```
titzer, line 155: Can we split this into two steps? One issue is that this method has the same name as a subroutine used here, but this one also performs a side effect in case of suspension.

titzer (approved):

PR #324, opened 2025-02-25
Add support for longer (14 bit) opcodes.

eliotmoss:
Does not yet include the run-time dispatch table building and use in the Fast interpreter, but builds and passes the tests I can run (I'm not set up for JVM tests at present).

### DIFF
--- a/src/engine/Module.v3
+++ b/src/engine/Module.v3
@@ -143,7 +143,7 @@ class FuncDecl(sig_index: int) extends Decl {
 		if (pc == 0) return void(entry_probed = true); // special case for function entry
 		// "orig" will become a copy of the original code, to allow in-place modification of old code
 		if (cur_bytecode == orig_bytecode) orig_bytecode = Arrays.dup(orig_bytecode);
-		cur_bytecode[pc] = u8.view(InternalOpcode.PROBE.code);
+		cur_bytecode[pc] = probe_byte;
 	}
 	def deactiveProbingAt(pc: int) {
 		if (pc == 0) return;
--- a/src/engine/Module.v3
+++ b/src/engine/Module.v3
@@ -143,7 +143,7 @@ class FuncDecl(sig_index: int) extends Decl {
 		if (pc == 0) return void(entry_probed = true); // special case for function entry
 		// "orig" will become a copy of the original code, to allow in-place modification of old code
 		if (cur_bytecode == orig_bytecode) orig_bytecode = Arrays.dup(orig_bytecode);
-		cur_bytecode[pc] = probe_byte;
+		cur_bytecode[pc] = u8.view(InternalOpcode.PROBE.code);
 	}
 	def deactiveProbingAt(pc: int) {
 		if (pc == 0) return;
--- a/src/engine/Opcodes.v3
+++ b/src/engine/Opcodes.v3
@@ -6,7 +6,7 @@ def sig_v_v = SigCache.v_v; // XXX: needed to force initialization
 def imm: ImmSigs;
 // An enumeration of the WebAssembly opcodes, including their mnenomic names,
 // the kind of immediate expected, and the (monomorphic) operator signature.
-enum Opcode(prefix: u8, code: u8, mnemonic: string, imms: Array<ImmKind>, sig: SigDecl) {
+enum Opcode(prefix: u8, code: u16, mnemonic: string, imms: Array<ImmKind>, sig: SigDecl) {
 	// Default, invalid opcode.
 	INVALID			(0x00, 0xFF, "<invalid>", imm.NONE, null),
 	// Control and calls.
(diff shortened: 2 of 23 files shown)


src/engine/Module.v3
```diff
@@ -143,7 +143,7 @@ class FuncDecl(sig_index: int) extends Decl {
 		if (pc == 0) return void(entry_probed = true); // special case for function entry
 		// "orig" will become a copy of the original code, to allow in-place modification of old code
 		if (cur_bytecode == orig_bytecode) orig_bytecode = Arrays.dup(orig_bytecode);
-		cur_bytecode[pc] = probe_byte;
```
titzer, line 146: Looks like a merge issue; the `probe_byte` is needed here because there is also `WHAMM_PROBE` as a possibility.

test/unittest/BinParserTest.v3
```diff
@@ -92,6 +92,11 @@ def Z = void(
 	()
 );
 
+// provide short forms
+def op8  (code: u16) -> u8 { return Opcodes.op8  (code); }
```
titzer, line 96: You can also simplify this to `def op8 = Opcodes.op8;`, here and in other files.

eliotmoss: Next commit addresses this.

src/engine/Module.v3
eliotmoss: Next commit addresses this.

src/engine/Opcodes.v3
```diff
@@ -727,30 +770,24 @@ def countOpcodes() -> i31 {
 component Opcodes {
 	def count = countOpcodes();
 	def page_by_prefix = Array<OpcodePage>.new(256);
-	def opcode_by_prefix = Array<Opcode>.new(256);
+	// def opcode_by_prefix = Array<Opcode>.new(256);
```
titzer, line 773: I think I'd like to keep the `opcode_by_prefix` name for this field, because it makes it a little more explicit at usage sites.

src/engine/Opcodes.v3
```diff
@@ -764,22 +801,47 @@ component Opcodes {
...
+		if (opcode.mnemonic.length > longestName) longestName = opcode.mnemonic.length;
+		if (opcode.prefix == 0) {
+			// Record non-prefixed opcode
+			var c = opcode.tag;
+			attributes[c] |= OpcodeAttribute.VALID;
+			if (opcode.sig != null && opcode.imms.length == 0) attributes[c] |= OpcodeAttribute.SHORT_OP;
+			opcodes[op8(opcode.code)] = opcode;
+			return;
+		}
+		// attributes[opcode.prefix] |= OpcodeAttribute.PREFIX;
```
titzer, line 814: Do we still need this line?

src/engine/Opcodes.v3
```diff
@@ -824,14 +886,42 @@ component Opcodes {
 		}
 		return Opcode.INVALID;
 	}
+	def subpage_added () {
+		++num_subpages;
+	}
+
+	def op8(code: u16) -> u8 {
```
titzer, line 893: Actually, now I see this, I think I'd prefer a `u8.!` at the usage sites; it fails under the same conditions and it will be optimized by the compiler.

titzer (commented):
Final comments, looks good otherwise.

src/engine/Opcodes.v3
eliotmoss: I get `page_by_prefix` since that array is indeed indexed by prefix.  But `opcode_by_prefix` does not have anything to do with prefixes - it's for opcodes that _don't_ have a prefix.  Maybe something like `opcodes_no_prefix`?

src/engine/Opcodes.v3
eliotmoss: Why change the uses as opposed to changing the body here to `u8.!(code)`?  I can change the other routines in that group similarly, though to force failure on `op16l` (for example) when its argument fits in 8 bits I can keep the division by 0 or do something like `u8.!(0x100 + code)`.  But if you prefer `u8.!(...)` to `op8(...)`, sure.  Thinking back, I believe I was trying to make what was going on a little more abstract, and uniform (i.e., `op8` as analogous to `op16l` and `op16h`).  Not really being argumentative here - just reaching for the best approach.

eliotmoss: I also see that `op14l` and `op14h` should make sure the code fits in 14 bits, perhaps using `u14.!` to filter the argument.  Is `u7.?` a reasonable test for a code to lie within 7 bits?  These are not used much w.r.t. opcodes, though it might be good to revisit usage of these and `op16l` and `op16h`, since in some cases a prefix allows a single byte 8-bit opcode, whereas in others a code in the range `0x80` to `0xFF` would require leb encoding.  If has to do with whtehr the opcode page is marked `oneByte`.

titzer: Yeah, I think `u8.!` inlined at the (many) uses puts my mind at ease and I'm fine with keeping the other routines since they are more involved. A `u14.!` will do the trick and `u7.?` will indeed do the proper unsigned value-range check.

titzer (approved):
lgtm; thanks!

titzer:
Oops, looks like I added a test in the meantime.

eliotmoss:
I don't understand the CI failure.  Looks like it is not using what I committed ...

titzer:
> I don't understand the CI failure. Looks like it is not using what I committed ...

Looks like it was a merge issue and the function got duplicated, with only one of the copies getting updated with the cast.

eliotmoss:
But is there an action I should take?  If so, what?  I've looked at the file as it was committed, and it seems fine (only one copy of the function, with the `u8.!` casts).

titzer:
I think you need to rebase your PR because the automatic merge isn't working.

PR #350, opened 2025-03-20
[stack-switching] Error handling with cross-stack unwind instead of `%ret_throw` register

linxuanm:
This PR extends stack walk to also unwind the parent stacks so that exception handling can be done without `%ret_throw`.


src/engine/x86-64/X86_64Stack.v3
```diff
@@ -1071,3 +1073,5 @@ component X86_64StackManager {
 		return result;
 	}
 }
+
+type StackWalkInfo(stack: X86_64Stack, frame: TargetFrame);
```
titzer, line 1077: We'll need to make this `#unboxed` to avoid an allocation for every frame.

linxuanm:
One thing I'm hesitant about is how `X86_64Runtime.runtime_THROW_DIRECT` is used here. What I'm trying to do is to create an exception object (which requires a runtime call), then throw the exception on the child stack without switching to the child stack's context (e.g., `vsp` and other interpreter registers). I think not switching to the child stack's is important as if the child stack is a freshly created continuation, then `resume_throw` can be executed without going through the child's `stack-enter-func` stub. Therefore the current solution is to call `X86_64Runtime.runtime_THROW_DIRECT` from the parent's context but with `rsp` and `curStack` set to the child, but this seems quite messy. Please let me know what you think!

src/engine/compiler/MacroAssembler.v3
```diff
@@ -290,6 +290,8 @@ class MacroAssembler(valuerep: Tagging, regConfig: RegConfig) {
 	def emit_call_runtime_callHost(func_arg: Reg);
 	def emit_jump_to_trap_at(reason: TrapReason);
 	def emit_call_runtime_op(op: Opcode);
+	def emit_get_curstack(r: Reg);
```
titzer, line 293: I recently started refactoring `MacroAssembler` to have explicit methods per field access, e.g. `emit_v3_Array_elem_r_rr` and so on. I think this will be cleaner if we also follow that convention here.

Those methods are now used exclusively in the JIT, and I think we should refactor the interpreter to follow suit.

src/engine/compiler/MacroAssembler.v3
```diff
@@ -302,6 +304,24 @@ class MacroAssembler(valuerep: Tagging, regConfig: RegConfig) {
...
+	// This should be called after all stack value transfers, but before the
+	// return address is pushed onto parent's {sp}.
+	// Chains {cont} on {parent} by:
+	// - setting the state of both {parent} and {cont}
+	// - setting {cont.parent} and {cont.parent_rsp_ptr}
+	// Destructive on {parent}.
+	def emit_chain_cont_to_parent(parent: Reg, cont: Reg);
+	// Called as the last step during resume. Switches {curStack} to the top of {cont}.
+	def emit_switch_curStack_to_cont(cont: Reg);
+	def emit_set_rsp_vsp(stack: Reg, rsp: Reg, vsp: Reg);
```
titzer, line 322: I didn't see why these combine two reads and writes. Can we make them separate and following the convention I mentioned above?

src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2422,51 +2422,81 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
...
 			var resumeStub = X86_64Label.new();
-			var cont_id = r_tmp0, r_ip_store = r_tmp1;
+			var cont_id = xenv.tmp0, ip_store = xenv.tmp1;
 			computeCurIpForTrap(-1);
 			computePcFromCurIp();
 			// read info and rewind %ip (need %ip at start of opcode during call)
-			asm.movq_r_r(r_ip_store, r_ip);
-			genReadUleb32(cont_id); // read signature index
-			asm.movq_r_r(r_ip, r_ip_store);
+			masm.emit_mov_r_r(ValueKind.REF, ip_store, xenv.ip);
```
titzer, line 2429: I don't quite understand why you need to store the `ip` from before the read of the continuation index.

src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2422,51 +2422,81 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
...
+
+			var cont = xenv.tmp1; // reuse due to non-overlapping live ranges
+			genPopInto(G(cont));
+			masm.emit_validate_and_consume_cont(xenv.tmp1);
+
 			// move parameter values from %curStack to %cont.top
-			var r_nvals = cont_id;
-			masm.emit_cont_mv(r_vsp, r_cont, r_nvals, r_tmp3, r_tmp4, r_xmm0);
+			var r_nvals = cont_id; // reuse due to non-overlapping live ranges
+			masm.emit_mov_r_m(ValueKind.REF, xenv.scratch, MasmAddr(xenv.instance, offsets.Instance_heaptypes));
```
titzer, line 2440: Following on above, we should have a `MacroAssembler` method for each field access; basically, `V3Offsets` should be used there.

src/engine/x86-64/X86_64Runtime.v3
```diff
@@ -234,6 +223,33 @@ component X86_64Runtime {
 		stack.rsp += -Pointer.SIZE;
 		return null;
 	}
+	// Used exclusively for {resume_throw} to throw without overwriting %rsp.
+	// This function is hacky in that it is called with {curStack.parent}'s
+	// context (to prevent transferring control flow to {curStack}), but uses the
+	// {rsp} of {curStack} to ensure stack unwinding works correctly.
+	def runtime_THROW_DIRECT(instance: Instance, tag_index: u31) -> Throwable {
```
titzer, line 230: This sounds a little dangerous and I think we should find a more robust solution. In particular, I don't understand what's special about a the parent of the stack that is doing a `resume_throw`. It seems like this should take the target continuation.

src/engine/x86-64/X86_64Interpreter.v3
linxuanm: I think this is because when handling a suspension `findSuspensionHandler` uses the parent stack's `ip` (which is excepted to point to the beginning of `resume`) to find the corresponding handler. In `V3Interpreter` this is done with a field that records the previous `resume`'s instruction pointer.

src/engine/compiler/MacroAssembler.v3
linxuanm: Can the explicit methods for field access be used for setting fields? I think currently there are only getters in the `MacroAssembler` refactor.

src/engine/x86-64/X86_64Interpreter.v3
titzer: This lookup should be based on the `pc` stored in the frame, not on the direct instruction pointer, which is a pointer into the middle of the bytecode.

PR #360, opened 2025-04-11
add a call graph monitor and fix some bugs

AllinLeeYL:
Dear developer:

I have recently tried this project. I want to say "Great Job!". I am trying to install this engine on MacOS, and another relatively outdated OS, CentOS 7. During installation, I found the installation not smooth because of the usage of the "progress" command and the argument "-e" for the "cat" command. It seems that the script cannot decide whether to use the progress command properly. Instead of introducing complex logic to decide whether progress is to be used, I decided to disable the "progress" command by default. Also, the "-e" argument does not work on some versions of "cat". It took me some time to locate the bugs and solve them. I made these changes and hope they will improve this project's portability.

Besides, I implemented a monitor to extract the call graph. The output will be in the format of "dot." I also hope this monitor can be merged into the repo. 

Best regards,


test/wasm-spec/update.sh
```diff
@@ -17,7 +17,7 @@ fi
 
 for p in $PROPOSALS; do
     printf "Updating proposal ${CYAN}%-22s${NORM} " $p
-    update_proposal_repo $p | $PROGRESS || exit $?
```
titzer, line 20: I'd like to keep these lines as they make the output of running these more concise.

src/monitors/CallGraphMonitor.v3
```diff
@@ -0,0 +1,119 @@
...
+// See LICENSE for details of Apache 2.0 license.
+
+// A general-purpose profiler.
+def monitor_ = MonitorRegistry.add(
+	"callgraph", "Extract dynamic call graph and output in dot format.",
+	CallGraphMonitor.new());
+def depthOpt = monitor_.options.newIntOption("depth", -1, "limit the maximum call tree depth");
+def filterOpt = monitor_.options.newDeclFilterOption("calls", "apply profiling to selected functions");
+
+class CallGraphMonitor extends Monitor {
```
titzer, line 11: Thanks for writing this monitor! This is cool functionality to have. The main feedback I have is that it collects similar information to both the `calls` monitor and the `profile` monitor. Of those two, maybe it makes sense to merge the dot graph functionality into the latter?

src/monitors/ProfileMonitor.v3
```diff
@@ -124,6 +134,55 @@ private class ProfileTreeNode(parent: ProfileTreeNode, func: WasmFunction) {
 		out.endColors();
 		out.ln();
 	}
+
+	def renderGraph(out: TraceBuilder, depth: int, maxDepth: int) {
+		if (func != null) {
+			var b = StringBuilder.new();
+			printFuncName(func, b);
```
titzer, line 141: You should be able to pass in the `out` directly into the `printFuncName` here.

AllinLeeYL:
I made some modification to report the execution time percentage in the dot graph. As the ProfileMonitor reports the dynamic execution time, some golden traces in `test/monitors/expected` are deleted.

titzer (approved):

PR #362, opened 2025-04-11
Fix the computation/flush of whamm br-mon

ejrgilbert:


src/monitors/WhammMonitor.v3
```diff
@@ -189,6 +189,31 @@ class WhammMonitor(named_modules: Array<NamedModule>) extends Monitor {
 						return null;
 					}
 				}
+				NumTargets     => {
+					match (op) {
```
titzer, line 193: This looks like a good candidate to factor out to a helper function.

titzer (approved):
lgtm other than small refactoring

PR #377, opened 2025-04-30
Various items necessary to fix the Whamm engine extension

ejrgilbert:
1. Fix out of bounds exception for check on " null" opcode variation of `ref.test` and `ref.cast`
2. Print stack trace on `throw`
3. Fix reverse order of `argN` handling (was following convention of `(call $target (i32.const 0) (i32.const 1))` which visually implies that the constants 0 and 1 are `arg0` and `arg1` respectively when it's actually the reverse)
4. Add new request variable support for the br_table's `num_targets` and a memory operation's `offset`
5. Fix `br-mon` flushing/computation (wasn't working before)


src/engine/Opcodes.v3
```diff
@@ -892,12 +892,12 @@ component Opcodes {
 			// so we check those opcodes to see if it has null at the end
 			match (op) {
 				REF_TEST => {
-					if (str.length >= p + 4 && Ranges.equal(str[p ... (p + 5)], " null")) {
+					if (str.length > p + 4 && Ranges.equal(str[p ... (p + 5)], " null")) {
```
titzer, line 895: You can do `[p ..+ 5]` for the range here.

src/monitors/WhammMonitor.v3
```diff
@@ -347,6 +354,56 @@ class WhammMonitor(named_modules: Array<NamedModule>) extends Monitor {
 		}
 		return args;
 	}
+	private def getNumTargets(op: Opcode, cv: CodeValidator, expType: ValueType) -> WhammArg {
+		var arg: WhammArg;
+		match (op) {
```
titzer, line 359: I think this will be a little less nested if you first check `op != Opcode.BR_TABLE` and return early.

src/monitors/WhammMonitor.v3
```diff
@@ -347,6 +354,56 @@ class WhammMonitor(named_modules: Array<NamedModule>) extends Monitor {
...
+		}
+		return arg;
+	}
+	private def getOffset(op: Opcode, cv: CodeValidator, expType: ValueType) -> WhammArg {
+		def imms = op.imms;
+		def parser = dupParser(cv);
+		var arg: WhammArg;
+
+		var set = false;
+		for (idx < imms.length) {
```
titzer, line 388: This is a little rickety; it looks like it won't work if the MEMARG is not the first immediate. There's also a growing amount of Whamm-specific bytecode decoding going on that could probably benefit from factoring into a more common place.

src/util/TraceUtil.v3
```diff
@@ -34,6 +34,29 @@ component TraceUtil {
 		out.putc('s');
 		return length + 1;
 	}
+	def renderThrowable(out: TraceBuilder, thrown: Throwable) {
```
titzer, line 37: I think `out` can be of type `StringBuilder` for a little more generality. Also, if it returns the `StringBuilder` then it will work with `%q` in string formatting.

titzer (commented):
Looks good with some suggestions.

titzer (approved):

PR #383, opened 2025-05-05
Rebasing stack-switching onto new `%sp` and stack walking convention

linxuanm:
This PR aims to refactor stack-switching in fast-int to accomodate changes to `%sp` and stack walking (specifically focusing on `resume_throw` and `throw/trap`):
- More tests for `resume_throw` (caught some bugs in `V3Interpreter`)
- ~Stack walk continues onto parent stack (trapping/throwing across stacks no longer uses `%ret_throw`, making it more consistent with throwable runtime calls)~ Maybe move this to another PR? This requires refactoring of `WasmStack` handling in general.
- Fixed `resume_throw` implementation in fast-int (there was a bug in exception handling in suspended continuation)
- Misc cleanup, such as refactoring stack-switching related code to use the new MacroAssembler V3 offset convention


src/engine/compiler/MacroAssembler.v3
```diff
@@ -311,6 +323,27 @@ class MacroAssembler(valuerep: Tagging, regConfig: RegConfig) {
 
 	def emit_debugger_breakpoint();
 
+	// Gets the number of parameters to a continuation.
```
titzer, line 326: I think I'd prefer to see this one inlined at the call sites, otherwise there is a lot of jumping backing and forth to the Masm.

src/engine/v3/V3Interpreter.v3
```diff
@@ -15,12 +15,11 @@ class V3Interpreter extends WasmStack {
...
-	// After `resume` is called, this stack is marked as CALL_CHILD and moves to
-	// the next instruction when resumed in the future. However at that time,
-	// the `pc` would point to the next instruction, making handler lookup
-	// impossible without somehow retrieving the `pc` at the `resume` that
-	// caused the CALL_CHILD transition. This variable is a hack for that.
-	private var prev_resume_pc = -1;
+	// XXX: remove this
+	// This flag is used for incrementing past a {resume} or {switch} opcode after
+	// being resumed. Suspending/switching shouldn't increment {frame.pc} until
+	// {resume} as {resume_throw} uses the old {pc} for error handling.
```
titzer, line 21: Is it possible to check for all errors before incrementing the PC? IIUC after suspending, the stack is left in a resumable state with the `pc` at the next instruction. Is it true the only error that can happen after calling the runtime to search for a handler that no handler is found? Then it seems that should be possible to handle in the caller and either update the `pc` before switching to the new stack, or trapping.

src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2440,54 +2440,76 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
...
+				// load new stack pointers
+				masm.emit_get_curstack(curStack);
+				masm.emit_v3_X86_64Stack_vsp_r_r(xenv.vsp, curStack);
+				masm.emit_v3_X86_64Stack_rsp_r_r(xenv.sp, curStack);
+				// pop address from new stack and go to it
+				masm.emit_pop_r(ValueKind.REF, xenv.scratch);
+				masm.emit_jump_r(xenv.scratch);
+			}
 		}
+		// bindHandler(Opcode.RESUME_THROW); {
```
titzer, line 2485: Is this code dead now?

src/engine/v3/V3Interpreter.v3
linxuanm: Sorry I don’t really understand. I was thinking that after a stack suspends, its pc points to the next instruction being transferring to the parent stack. Therefore if resume_throw is called on this suspended stack, the throw pc will be misplaced (the expected throw pc of a resume_throw should be the pc of the suspend/switching instruction that suspended the stack). Are you suggesting that this adjustment should be done by the parent that called resume_throw?

src/engine/x86-64/X86_64Interpreter.v3
linxuanm: There was a bug in the old implementation that causes new tests to fail. I plan to implement the new stack walk first and base this handler on top of that in a separate PR (since the handler would change a lot after the rewrite) instead of fixing the current implementation. What do you think?

src/engine/v3/V3Interpreter.v3
titzer: Ah, right, hadn't thought about that. What about storing the suspending `pc` separately? That's maybe a similar amount of complexity but less subtle. It shouldn't make any difference for finding a handler in `resume_throw` because before or after the suspend will be in the same control scope (and thus have the same handlers).

linxuanm: I think this differs from the reference interpreter behavior (e.g., on [test/regress/ext:stack-switching/resume_throw9.wast](test/regress/ext:stack-switching/resume_throw9.wast), where the tag thrown by `resume_throw` is supposed to be caught by the `try_table`).

titzer: AFAICT we could fix that by considering the `end` of that `try_table` to be inside the `try_table`. We should also develop regular EH tests for such boundary cases.

src/engine/x86-64/X86_64Interpreter.v3
titzer: Sounds ok.

PR #394, opened 2025-06-09
Only perform calls in match rules if other params result in a match

ejrgilbert:
There is a bug if we have the following:

```
(func "wasm:opcode:i32.const  ($15(fid, pc), local0)" (param i32 i32) ...)
```

This probe has a type bound on `local0`. We only have a match if the `local0` is, in fact, an `i32`. If it's not an `i32`, we don't have a match and the probe won't be inserted. This logic is done correctly, the probe isn't inserted...but the static allocation function is still called during parameter parsing. This means that we allocate memory for a probe that we'll never use since it's not inserted.

This PR fixes this issue through lazily performing these calls if other parameter preparation results in a match.


src/util/Whamm.v3
```diff
@@ -302,14 +304,26 @@ component WhammVarBinder {
 						err.at(cv.parser.decoder).WhammProbeError("whamm probe", "nested function calls");
 						return null;
 					}
-					arg = bindCallResult(target, params, wi, op, expParams[i], cv, err);
+					// Call this function only if we have a match!
```
titzer, line 307: This might be a bit shorter if you use a list of funexprs, so that you don't need to define an intermediate data structure.

PR #395, opened 2025-06-10
Only perform calls in match rules if other params result in a match

ejrgilbert:
There is a bug if we have the following:

```
(func "wasm:opcode:i32.const  ($15(fid, pc), local0)" (param i32 i32) ...)
```

This probe has a type bound on `local0`. We only have a match if the `local0` is, in fact, an `i32`. If it's not an `i32`, we don't have a match and the probe won't be inserted. This logic is done correctly, the probe isn't inserted...but the static allocation function is still called during parameter parsing. This means that we allocate memory for a probe that we'll never use since it's not inserted.

This PR fixes this issue through lazily performing these calls if other parameter preparation results in a match.


titzer:
My comments were minor; I'll land this and make small adjustments.

PR #398, opened 2025-06-13
Support requesting `fname` in Whamm calls at bind time

ejrgilbert:
Note: This does not work when `fname` is requested as a parameter to the probe callback. To support this, we would need to create a new WhammArg that encodes the bytes to write to memory on a probe callback invocation or something of this nature.

This is just the design I thought up, I'm open for suggestions/design change!


titzer:
Hmm, I scanned through the implementation and was thinking that maybe we could require bindings of `fname` to match up with a match time callback that gets called to allocate space for the funcname?

```
(export "$store_fname" (func (param i32) (result i32 i32) ...) // malloc()'s memory to store fname, given the size, and returns its address and size)
(export "wasm:opcode:call($store_fname(fname))" (func (param i32 i32)) ...) // probe gets the function name address and size bound by the match-time callback

ejrgilbert:
> Hmm, I scanned through the implementation and was thinking that maybe we could require bindings of `fname` to match up with a match time callback that gets called to allocate space for the funcname?
> 
> ```
> (export "$store_fname" (func (param i32) (result i32 i32) ...) // malloc()'s memory to store fname, given the size, and returns its address and size)
> (export "wasm:opcode:call($store_fname(fname))" (func (param i32 i32)) ...) // probe gets the function name address and size bound by the match-time callback
> ```

This would work for supporting the dynamic usage, I'll add that. It shouldn't change anything about how it's used statically though. Do you also agree with having the engine write to some memory that's identified through the export name "engine:data"?

PR #411, opened 2025-06-30
Add abstract interpreter to GRP monitor

matthew-mojira:


titzer:
As discussed, all this needs is some test cases and it's good to go.

titzer:
@matthew-mojira Is this PR still active?

titzer:
Looks like the abstract interpreter is running out of memory when running on the Virgil interpreter. Is it leaking some information between different runs?

titzer (approved):

PR #424, opened 2025-07-19
Split CodeValidator.validateFunc to allow introspection at func entry

titzer:


titzer:
I think it's better just to have another probe callback.

PR #448, opened 2025-08-06
[stack-switching] Continuation handling & tests

linxuanm:
This PR adds more support for handling continuation storage in preparation for unboxed continuation.


linxuanm:
I'll also include the changes to `ValueKind` in this PR, and then let's see if we can land it!

PR #455, opened 2025-08-09
[CBD] Lots of CBDs, uniformly use i32

mkhan45:
- Lots of numerics, f32 stuff
- doIf, doTry
- Reference instructions
- i32 load/store
- Intrinsic file with stages and effects


titzer:
> When we talked about it before, we decided to prefer `i32` since that's what the WASM spec does. I think `u32` makes more sense though.
> 
> To clarify, if CBD only has signed ints, we have to pass i32s to the intrinsics and expect their implementations to `u32.view` them. OTOH, if CBD only has unsigned ints, we should be able to just `push_u32` in `i32.const` without having to `i32.view` anything except for in signed operations, right?
> 
> The force push was for the rebase

Yeah, I went through similar reasoning in Wizard and essentially arrived at: `u32` is pervasive in the format (as LEBs) and `u32` is slightly more efficient or ergonomic for doing some bit-level things. Signedness only shows up in a few places. If you notice in Wizard, most of the immediate LEB indices and such are actually `u31` since they are subject to the standard format limitations (e.g. `func_index` by the number of functions) but Virgil requires `int` in various places. Since a `u31` promotes to `int`, it was relatively nice.

PR #457, opened 2025-08-13
[wali] Refactor WALI implementation to use new HostModuleBuilder .func_xxx_x

titzer:
@MSMazaya PTAL.

We've added a new mechanism for exposing host modules that unboxes the Wasm arguments to Virgil arguments. That simplifies host modules and instances a lot, but necessitated a refactor to the WALI implementation.


MSMazaya:
Neat. Thanks for tagging me and sorry for not having done the refactor myself. I will be back to PGH by Aug 22 and will continue to work on it.

PR #468, opened 2025-09-01
Add `new_func` to wave target

matthew-mojira:


src/engine/Instance.v3
```diff
@@ -79,7 +79,7 @@ class ExportedValueType(src: ExportedType, vt: ValueType) extends ExportedType {
 // An instantiated module, including memories, tables, functions, etc.
 class Instance(module: Module, imports: Array<Exportable>) {
 	def memories = Array<Memory>.new(module.memories.length);
-	def functions = Array<Function>.new(module.functions.length);
+	var functions = Array<Function>.new(module.functions.length);
```
titzer, line 82: The prior function in `wizeng.new_funcref` doesn't append to the instance. I think we can get by without appending for now. That means that these functions can be garbage-collected.

src/modules/wave/WaveModule.v3
```diff
@@ -62,6 +68,46 @@ class WaveInstance {
...
+
+		var func: FuncDecl = FuncDecl.new(sig_idx);
+		func.setOrigCode(code);
+		func.sig = sig;
+		func.func_index = instance.functions.length; // but not yet added
+
+		/* validate */
+		var err: ErrorGen = ErrorGen.new(instance.module.filename);
+		Execute.tiering.onFuncBody(instance.module, u32.!(func.func_index), code, err);
+		var validator = CodeValidator.new(Extension.set.all, Limits.new(), instance.module, err);
```
titzer, line 92: Let's cache a lazily-created code validator in the WaveInstance to avoid constructing one each time.

titzer (approved):
lgtm; very nice

titzer:
> lgtm; very nice

very very nice

PR #471, opened 2025-09-01
[wizeng] Merging `spectest` into `wizeng` executable

linxuanm:
This PR removes the `spectest` target and moves the testing functionality into the main `wizeng` target.

## Behavior Description
Instead of having two executables, when running the command:
```
bin/wizeng <options> [<*.wasm>|<*.bin.wast>]
```
`wizeng` will first execute the `*.wasm` files normally, then run the tests in each of the `*.bin.wast` file. This eliminates the need for extra flags/convoluted execution mode separation (design was discussed in-person).

The main purpose of this PR is to unify the engine setup for `spectest` and `wizeng` so that monitors work when running spectests.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -38,8 +38,9 @@ TARGET_V3="src/engine/v3/*.v3"
 TARGET_X86_64="src/engine/compiler/*.v3 src/engine/x86-64/*.v3 $VIRGIL_LIB/asm/x86-64/*.v3"
 UNITTEST="test/unittest/*.v3 test/wasm-spec/*.v3 test/unittest.main.v3 $VIRGIL_LIB/test/*.v3"
 UNITTEST_X86_64_LINUX="test/unittest/x86-64-linux/*.v3"
-SPECTEST="test/wasm-spec/*.v3"
-WIZENG="src/SpectestMode.v3 src/WasmMode.v3 src/wizeng.main.v3 src/modules/*.v3 src/modules/wizeng/*.v3"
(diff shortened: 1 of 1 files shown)


src/SpectestMode.v3
```diff
@@ -0,0 +1,71 @@
+// Copyright 2020 Ben L. Titzer. All rights reserved.
```
titzer, line 1: Copyright should be "Wizard authors".

src/wizeng.main.v3
```diff
@@ -35,239 +36,60 @@ def main(args: Array<string>) -> int {
...
-	for (i < WizengOptions.REPEAT.val) {
-		exit_code = linkAndExecuteMain(engine, monitors, t.1, programArgs);
-		if (exit_code != 0) break;
+	
+	if (wasmFiles.length > 0) {
+		exit_code = WasmMode.run(engine, monitors, wasmFiles, programArgs);
+		if (exit_code != 0) return exit_code;
+	}
+	if (wastFiles.length > 0) {
+		exit_code = SpectestMode.run(engine, monitors, wastFiles);
```
titzer, line 84: It'd be good to have a build option `--no-spec-test` which doesn't include the spec test runner at all (and maybe similarly `--no-wasm-run`).

One way to support that is to have a `var` field in this component that is a function for running spec tests, and have the `SpecTestMode` initialize that field to point to one of its functions (at compile time). That way, if the `SpecTestMode` code is not included at build time, the `var` field will be null; the `main` runner can then be adjusted accordingly.

PR #477, opened 2025-09-18
Use `#!/usr/bin/env bash` instead of `#!/bin/bash`

samestep:
[It's better](https://stackoverflow.com/a/10383546/5044950); for instance, `#!/bin/bash` doesn't work on NixOS. See also titzer/virgil#482.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 function exit_usage() {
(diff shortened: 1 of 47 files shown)


ejrgilbert:
I don't know that it's necessarily "better", most scripts use `#!/bin/bash`. Systems where that doesn't work could just create a symlink to fix the issue (which would fix the issue for the majority of scripts that use `#!/bin/bash`).

samestep:
@titzer any chance I could bribe you with a pop tart for this one too?

titzer:
I am open to landing this when the CI is green.

samestep:
Alrighty @titzer looks like it's green!

PR #478, opened 2025-09-18
[debug] Debugging guide + utils

linxuanm:
This PR adds some debugging related features and documentation. It also adds a `--debug` build in `build.sh` that enables the debug monitors (currently only one placeholder monitor used for next PR, as the `monitors/debug/` directory cannot be empty or `v3c` will throw an error about file not found).


titzer:
I think this needs to be rebased now; CI is green.

PR #516, opened 2025-10-16
[exception-handling] Fixed bug where unreachable handlers produce incorrect handler stub indices

linxuanm:
This PR fixes a bug where handler linking is inconsistent due to mismatched handler indices when the SPC skips over unreachable code.

### DIFF
--- a/src/engine/Module.v3
+++ b/src/engine/Module.v3
@@ -113,7 +113,7 @@ class Decl {
 def NO_EX_HANDLERS: Array<ExHandlerEntry> = [];
 def NO_DEST_MAP: Array<int> = [];
 def NO_DESTS: Array<TargetHandlerDest> = [];
-def NO_PC_MAP = HashMap<int, Vector<int>>.new(int.!, int.==);
+def NO_PC_MAP: Array<(int, Array<int>)> = [];
 class FuncHandlerInfo {
 	var ex_handlers = NO_EX_HANDLERS;
 	var suspend_handlers = NO_EX_HANDLERS;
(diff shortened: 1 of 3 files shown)


linxuanm:
I think there is room for further simplification of how the linking is done, now that we are using a `pc` lookup rather than relying on handler indices. Do you think we should land this for now to keep things going (I'm thinking of fixing the local variable bug first) and address that in the future?

src/engine/Module.v3
```diff
@@ -113,12 +113,14 @@ class Decl {
...
 def NO_DEST_MAP: Array<int> = [];
 def NO_DESTS: Array<TargetHandlerDest> = [];
+def NO_PC_MAP = HashMap<int, Vector<int>>.new(int.!, int.==);
 class FuncHandlerInfo {
 	var ex_handlers = NO_EX_HANDLERS;
 	var suspend_handlers = NO_EX_HANDLERS;
 	var switch_handlers = NO_EX_HANDLERS;
 	var handler_dest_map: Array<int> = NO_DEST_MAP;
 	var handler_dests: Array<TargetHandlerDest> = NO_DESTS;
+	var handler_pc_index_map = NO_PC_MAP;
```
titzer: I think we should also modify the `findHandler()` method to use this new datastructure.

linxuanm: I think it is a good idea to enforce the sorted array on all types of handlers so that `findHandler()` can do a binary search on each of them. I'm thinking instead of letting each `xxx_handlers` hold a handler entry, it should hold a handler index into the array of all handlers, and `handler_pc_index_map` would just be the array of all handlers in sorted order by `pc`. Without this change I don't think `findHandler` can use `handler_pc_index_map` as it is indexed by the source `pc` of the handler, not its range. What do you think?

linxuanm:
I just realized that we can't sort exception handlers by their opcode range because there might be `catch_all` handlers at the end of a handler table (or multiple handlers for the same tag in a nested block). Do you have suggestions on fixing this? Alternatively could we land this PR for now and implement a binary search later? I rolled-back this branch to a working linear search.

PR #531, opened 2025-11-29
[stack-switching] Value.Cont

linxuanm:
This PR adds another case for `Value` to store the continuation type, which can be built with `--unboxed-cont` flag.


linxuanm:
This is closed in favor of #547.

PR #607, opened 2026-04-11
ADT `X86_64HandlerFrame`

linxuanm:
This PR adds an ADT `X86_64HandlerFrame` that wraps field accesses to both the interpreter and the SPC stack frame.

### DIFF
--- a/src/engine/x86-64/X86_64Stack.v3
+++ b/src/engine/x86-64/X86_64Stack.v3
@@ -961,10 +961,7 @@ layout X86_64SpcFrame {
 	=104;
 }
 
-// Lightweight, discriminated handler for interpreter and SPC frames.
-// Unlike {X86_64FrameAccessor}, does not require heap-caching on the frame's accessor slot,
-// and does not store a FuncDecl/stack/etc. Use this for operations that just need
-// to read/write frame fields without full probe instrumentation support.
+// Frame handler for interpreter and SPC stack frames; a lightweight wrapper for frame slot accesses.
 type X86_64FrameHandler {
 	case Interpreter(sfp_: Pointer) {
 		def sfp() -> Pointer { return sfp_; }
@@ -1128,9 +1125,7 @@ type X86_64FrameHandler {
 	}
 
 	// Write SPC execution state fields needed when an SPC frame transitions to a new
-	// program location (e.g., jumping to an exception/suspension handler stub). Does NOT
-	// write interpreter-only fields, which would corrupt the SPC slot aliasing
-	// ({inlined_instance} at +40, {inlined_mem0_base} at +56) and has no consumer in SPC code.
+	// program location (e.g., jumping to an exception/suspension handler stub)
 	def writeSpcState(new_pc: int) {
 		set_curpc(new_pc);
 		set_accessor(null);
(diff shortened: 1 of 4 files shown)


src/engine/x86-64/X86_64Stack.v3
```diff
@@ -938,6 +942,230 @@ layout X86_64InterpreterFrame {
...
+	+56	inlined_mem0_base	: i64;	// Pointer — Whamm probe inlining (aliases interpreter ip slot)
+// +64: unused
+// +72: unused
+	+80	instance		: i64;	// Instance (shared with interpreter)
+	+88	curpc			: int;	// int — written on trap/inlined-frame paths for pc() lookup
+	+96	accessor		: i64;	// FrameAccessor (shared with interpreter)
+	=104;
+}
+
+// Frame handler for interpreter and SPC stack frames; a lightweight wrapper for frame slot accesses.
```
titzer: Let's name these "frame handles" (not handler).

src/engine/x86-64/X86_64Stack.v3
```diff
@@ -938,6 +942,230 @@ layout X86_64InterpreterFrame {
...
+type X86_64FrameHandler {
```
titzer: This needs to be unboxed because they should not be allocated on the heap, particularly because they could be used during GC.

src/engine/x86-64/X86_64Stack.v3
```diff
@@ -938,6 +942,230 @@ layout X86_64InterpreterFrame {
...
+	case Interpreter(sfp_: Pointer) {
```
titzer, line 969: I think this will be a lot cleaner if the constructor has a `Ref<X86_64InterpreterFrame>` layout. We can then define accessor methods in the type and override them for both the interpreter and the SPC frame case.

src/engine/x86-64/X86_64Stack.v3
```diff
@@ -938,6 +942,230 @@ layout X86_64InterpreterFrame {
...
+		(sfp() + X86_64InterpreterFrame.vsp.offset).store<Pointer>(val);
+	}
+	def set_curpc(val: int) {
+		(sfp() + X86_64InterpreterFrame.curpc.offset).store<int>(val);
+	}
+	def set_accessor(val: X86_64FrameAccessor) {
+		(sfp() + X86_64InterpreterFrame.accessor.offset).store<X86_64FrameAccessor>(val);
+	}
+
+	// Guards.
```
titzer: I think it's better to use Virgil's cast and query operators instead of wrapping them with additional indirection. So e.g. we can just do `X86_64InterpreterFrame.Interpreter.?()` or do a `match` in client code.

titzer (commented):
This is a good start! I like how much this cleans up client code. See suggestions below on how to improve the ADT itself.

src/engine/x86-64/X86_64Stack.v3
linxuanm: I think this might run into the problems of `layout X86_64InterpreterFrame`'s fields being primitives, e.g., we won't be able to extract `wasm_func` as a `WasmFunction` directly and instead need to do another pointer arithmetic and load because `Ref<>` uses the field of the type. Is there a way around this?

titzer: The `Pointer.atField` operator should work with layout fields. This is preferred to pointer arithmetic. It will still require an unsafe load, e.g.

```
 type X86_64FrameHandle {
   case Spc(r: Ref<X86_64SpcFrame>) {
     def wasm_func() => Pointer.atField(r.wasm_func).load<WasmFunction>();
     ...
   }
   case Interpreter(r: Ref<S86_64InterpreterFrame>) {
    def wasm_func() => Pointer.atField(r.wasm_func).load<WasmFunction>();
     ...
   }
}
```

PR #609, opened 2026-04-15
Add check for "entry" on wave

matthew-mojira:
- [x] bump interpreter code size
- [x] add tests


matthew-mojira:
I don't see that files in `wave` are tested. Is this supposed to be a source folder for tests elsewhere or its own testing set?

titzer:
> I don't see that files in `wave` are tested. Is this supposed to be a source folder for tests elsewhere or its own testing set?

I think it's missing a test script (oops).

src/modules/wave/WaveModule.v3
```diff
@@ -48,7 +48,12 @@ class WaveInstance {
 	}
 	// Get the main function of an instance, if there is one.
 	def getMain(instance: Instance) -> (Function, Array<Value>) {
-		var main = instance.findExportOfType<Function>(GlobMatcher.new("main"));
+		// First: look for function called entry
+		var main = instance.findExportOfType<Function>(GlobMatcher.new("entry"));
+		// If not: look for function called main
+		if (main == null)
```
titzer: Why not just `if (main != null) return main;` ?

titzer (commented):
Minor suggestion

titzer (approved):
lgtm

PR #614, opened 2026-04-24
Move save and restore of interpreter registers to MacroAssembler

matthew-mojira:


src/engine/x86-64/X86_64MacroAssembler.v3
```diff
@@ -21,9 +21,28 @@ class X86_64MacroAssembler extends MacroAssembler {
 	var scratch: X86_64Gpr;
 	var jump_tables: Vector<(int, Array<X86_64Label>)>;
 	var trap_stubs: X86_64SpcTrapsStub;
+	var all_ivars: Array<(Reg, int)>;
```
titzer: I think this should be part of the `ExecEnv`; that way it can be cached instead of reconstructed for every MacroAssembler creation.

titzer (approved):
lgtm

PR #617, opened 2026-04-28
Call `unrefReg` at the end of stack reconstruction

matthew-mojira:
and move the `unrefReg` before to be inside the method


titzer:
Can you add a regression test that this fixes?

titzer (approved):

titzer:
As discussed, for fast handlers, this can be taken care of by pre-allocating the registers used by the interpreter and letting the register allocator spill them.

PR #621, opened 2026-05-06
[stack-switching] Restore PC after resume finishes

linxuanm:
This PR fixes a bug that occurs due to the PC not resetting properly after resumption finishes.


titzer:
Can you provide a regression test for this one?

linxuanm:
I tried adding 3 regression tests for this (nested completed resumption in an attempt to rule out stale PC register) but they weren't able to fail without the fix. Can we maybe land this and the 3 new tests for now, and consider a monitor-based testing approach (for directly inspecting the PC after a stack returns) during tomorrow's meeting?

titzer (approved):

PR #627, opened 2026-05-10
Full support for `REF_U64`

linxuanm:
This PR adds full support for using `REF_U64` value kinds with unboxed continuations, including for null continuation values.


src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -2369,13 +2369,42 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
...
 		}
 		bindHandler(Opcode.REF_NULL); {
-			genSkipLeb();
-			genTagPush(BpTypeCode.REF_NULL.code);
-			asm.movq_m_i(vsph[0].value, 0);
-			// XXX: only clear upper slot when REF is guaranteed a REF_U64
-			asm.movq_m_i(vsph[0].value.plus(8), 0);
-			incrementVsp();
-			endHandler();
+			var l_is_cont = X86_64Label.new(), l_not_cont = X86_64Label.new();
```
titzer: What about the decoding of locals of type continuation? Do we also need to update the function entry sequence?

titzer (approved):
lgtm

PR #632, opened 2026-05-11
Add a debugger feature to dump all the globals.

ezrakilty:
As an enhancement, I'd like to add a feature to print a particular global. TBD.


titzer (approved):
lgtm, thanks!

PR #633, opened 2026-05-11
Add a debugger feature to print 64-bit words in memory.

ezrakilty:
This one is a little underbaked as yet, but I wanted to share the idea already.

It should have:
* The ability to print from other memories
* The ability to print i32, i64, u32, u64 types at least. (Reading little-endian int data out of hex digits is 🤡 )

But maybe you have some ideas what the interface should be for selecting those? I was thinking maybe `p [<memory>] [<type>] <address>` where <memory> is the index of the memory object, and would default to 0. And <type> could be e.g. `i32`, etc, or it could be `raw` or `hex` or something to indicate you want the raw bytes in hex. It would default to raw.

I must say, these two debugger features I'm PRing were _crucial_ to pinning down my vexing bug in the Go compiler today. I was finally able to inspect the shadow stack at each point and see where I was going wrong. :pray:

### DIFF
--- a/src/monitors/DebugMonitor.v3
+++ b/src/monitors/DebugMonitor.v3
@@ -39,6 +39,7 @@ type DbgCommand {
 	case Continue;
 	case List;
 	case Set(local_idx: int, value: Value);
+	case Print(index: int);
 	case Unknown(cmd: string);
 }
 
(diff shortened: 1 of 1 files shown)


titzer:
Nice! One idea would be to try to mirror the usage of the `x` command in `gdb` (https://visualgdb.com/gdbreference/commands/x). That'd be useful for people who have a `gdb` background to transfer skills. Wdyt?

ezrakilty:
Yeah, I would try to mimic gdb. I'll just have to refresh my memory what gdb actually does :sweat_smile: I'll get back to this!

PR #643, opened 2026-05-28
[stack-switching] Stricter value type reads

linxuanm:
This PR makes sure the read paths of `REF` in `X86_64Stack` reject continuation type due to it being specialized.


titzer (approved):

titzer:
I'll waive this through, but the real fix as discussed yesterday is to refactor casts.

PR #644, opened 2026-05-29
Runtime cast refactor

linxuanm:
This PR refactors `Runtime.cast` so that the `x86-64` paths performs runtime casting without allocating a `Value`.


titzer (approved):
lgtm; we can factor out some functionality and inline the branch into the interpreter and JITs as a next step

PR #645, opened 2026-05-30
Stack switch event monitor

linxuanm:
This PR replaces the old `ssinstr` monitor with a new event tracing monitor to trace all stack switching operations. It cleans up the old monitor's method of recording frequency-related statistics (through the histogram), so that visualization/analysis can be moved to the downstream Python script for data analysis.


src/monitors/StackSwitchEventMonitor.v3
```diff
@@ -0,0 +1,152 @@
...
+// Single global probe: fired on every executed instruction. Looks up the current PC in a
+// precomputed sparse per-function site map and, on a hit, appends the site (with the number
+// of instructions executed since the last stack-switching op) to the ordered event log.
+// Fast path (non-switch instruction) is one HashMap get + bounds + null check + increment.
+class EventDispatchProbe(monitor: StackSwitchEventMonitor) extends Probe {
+	var gap: u64; // instructions executed since the last stack-switching op
+
+	def fire(dynamicLoc: DynamicLoc) -> ProbeAction {
+		var fsites = monitor.funcSites[dynamicLoc.func.decl];
+		if (fsites != null && dynamicLoc.pc < fsites.length) {
```
titzer: If we use local probes for the stack switch events, then the global probe doesn't need to query for every site, and there isn't a need for a parallel metadata array.

src/monitors/StackSwitchEventMonitor.v3
```diff
@@ -0,0 +1,152 @@
...
+				return ProbeAction.Continue;
+			}
+		}
+		gap++;
+		return ProbeAction.Continue;
+	}
+}
+
+// Records per-PC EventSites for each function.
+class EventSiteVisitor extends StackSwitchSiteVisitor {
```
titzer: Since there is only one subclass, I think we should merge this into the instrumenter.

src/monitors/StackSwitchEventMonitor.v3
```diff
@@ -0,0 +1,152 @@
...
+	def monitor: StackSwitchEventMonitor;
+	var pcSites: Array<EventSite>;
```
titzer: This will go away with local probes.

titzer (approved):
lgtm

PR #646, opened 2026-06-03
Fix incorrect trap reason with negative call_indirect index

matthew-mojira:


titzer (commented):
Can you add a regression test as well? This is nice to verify the behavior against the spec interpreter.

titzer (approved):

PR #647, opened 2026-06-06
x86-64 Stack Compression

linxuanm:
Stack compression backend for `x86-64`.

### DIFF
--- a/src/engine/compression/Compression.v3
+++ b/src/engine/compression/Compression.v3
@@ -6,8 +6,9 @@
 type WasmFrameData(func: WasmFunction, pc: int, ret_addr: u64, vals: Array<Value>) #unboxed { }
 
 // Relocatable representation of a stack frame that can be serialized.
-// {bytecode_ip} is unused when {is_spc} is true.
-type RelocatableFrame(data: WasmFrameData, is_spc: bool, bytecode_ip: u64) #unboxed { }
+// {bytecode_offset} is the byte offset into {data.func.decl.cur_bytecode} at which the interpreter
+// resumes (unused when {is_spc} is true).
+type RelocatableFrame(data: WasmFrameData, is_spc: bool, bytecode_offset: int) #unboxed { }
 
 class CompressionStrategy<C> {
 	def compress(frames: Range<RelocatableFrame>) -> C;
--- a/src/engine/x86-64/X86_64Stack.v3
+++ b/src/engine/x86-64/X86_64Stack.v3
@@ -45,16 +45,23 @@ class X86_64Stack extends WasmStack {
 		return state_;
 	}
 	// Requires {state == EMPTY}.
-	// Resets this stack to be {SUSPENDED}, awaiting arguments for {func}.
-	def reset(func: Function) -> this {
+	def reset0(func: Function, remaining_params: int) -> this {
 		checkState("reset()", StackState.EMPTY);
 		this.func = func;
-		params_arity = func.sig.params.length;
+		params_arity = remaining_params;
 		return_results = func.sig.results;
 		state_ = if(params_arity == 0, StackState.RESUMABLE, StackState.SUSPENDED);
 		parent_rsp_ptr = pushRspPointer(Pointer.NULL); // placeholder for parent pointer
 		pushRspPointer(STACK_RETURN_PARENT_STUB.getEntry() + NOP_ENTRY);
+	}
+	// Requires {state == EMPTY}.
+	def reset(func: Function) -> this {
+		reset0(func, func.sig.params.length);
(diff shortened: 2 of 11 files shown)


src/engine/compression/Compression.v3
```diff
@@ -6,8 +6,9 @@
 type WasmFrameData(func: WasmFunction, pc: int, ret_addr: u64, vals: Array<Value>) #unboxed { }
 
 // Relocatable representation of a stack frame that can be serialized.
-// {bytecode_ip} is unused when {is_spc} is true.
-type RelocatableFrame(data: WasmFrameData, is_spc: bool, bytecode_ip: u64) #unboxed { }
+// {bytecode_offset} is the byte offset into {data.func.decl.cur_bytecode} at which the interpreter
+// resumes (unused when {is_spc} is true).
+type RelocatableFrame(data: WasmFrameData, is_spc: bool, bytecode_offset: int) #unboxed { }
```
titzer, line 11: Let's use "pc" instead of "bytecode_offset"--that matches the rest of the code.

src/engine/x86-64/X86_64Stack.v3
```diff
@@ -45,16 +45,19 @@ class X86_64Stack extends WasmStack {
 		return state_;
 	}
 	// Requires {state == EMPTY}.
-	// Resets this stack to be {SUSPENDED}, awaiting arguments for {func}.
-	def reset(func: Function) -> this {
+	def resetWithState(func: Function, pending_params: int, push_enter_stub: bool) -> this {
```
titzer: How about `remaining_params`?

src/engine/x86-64/X86_64Stack.v3
```diff
@@ -45,16 +45,19 @@ class X86_64Stack extends WasmStack {
 		return state_;
 	}
 	// Requires {state == EMPTY}.
-	// Resets this stack to be {SUSPENDED}, awaiting arguments for {func}.
-	def reset(func: Function) -> this {
```
titzer, line 49: I think we should keep this method and have it forward to the general routine and after that, push the return stub, rather than having a boolean parameter controlling it.

test/unittest/x86-64-linux/X86_64CompressionTest.v3
```diff
@@ -0,0 +1,732 @@
...
+def newObject = ObjectModel.newStructFromValues(EMPTY_STRUCT, _);
+
+def newIntTester(t: Tester) -> X86_64CompressionTester {
+	return X86_64CompressionTester.new(t, X86_64InterpreterOnlyStrategy.new());
+}
+def newSpcTester(t: Tester) -> X86_64CompressionTester {
+	return X86_64CompressionTester.new(t, X86_64SpcAotStrategy.new(false));
+}
+
+component X86_64CompressionTest {
```
titzer: Can you follow the pattern of the other unit tests and lift these to the top level?

test/unittest/x86-64-linux/X86_64CompressionTest.v3
```diff
@@ -0,0 +1,732 @@
...
+			("vals", test_spc_vals),
+			("pc_is_derived", test_spc_pc_is_derived),
+			("curpc", test_spc_curpc)
+			]) {
+			UnitTests.registerT("x86compress:spc:", test.0, newSpcTester, test.1);
+		}
+	}
+}
+
+class X86_64CompressionTester extends ExeTester {
```
titzer: Just make the class private so that you don't need to repeat that for every field.

test/unittest/SidetableTest.v3
```diff
@@ -267,4 +268,32 @@ def test_br_loop_val(t: SidetableTester) {
 		BR, 0,
 		END]);
 	t.assert(5, -4, 0, 1, 0);
-}
\ No newline at end of file
+}
+// {CodeValidator.refR} writes a {Sidetable_ResumeEntry} for every resume instruction, plus one
```
titzer: This big comments are a typical Claude-ism. It's better to have 1-2 lines max; there isn't anything that requires that much explanation in a test.

src/engine/compression/Compression.v3
linxuanm: I believe `bytecode_offset` is different from `pc` here. `bytecode_offset` was initially referring to `ip`, which differs from `pc` because a non-top stack in the chain has `pc` pointed at the start of its `resume` call while `ip` is advanced forward. It was initially named `bytecode_ip` to store the partial `ip` as a pointer casted to a `u64`, but this will fail if the `cur_bytecodes` has been moved by the GC and the `bytecode_ip` will point to recycled memory after a GC call. Due to this I changed it to store the offset of `ip` from `cur_bytecodes` instead. Should I name this to `ip_offset` or something?

titzer (approved):

PR #651, opened 2026-06-15
Add tests for fast handlers

matthew-mojira:
Fast handler tests all have expected failures on, since it's not implemented.

The fastcall tests are run on the `int` and `jit` modes. Unlike the other JIT-compiled modes (except maybe `dyn` when not hot enough), `jit` does not JIT-compile dynamically added functions (using `wave.new_func`). That is used (abused) for test 72, so that we can have a out call in a fast handler to an expensive JIT-compiled function. Should probably use filter for JIT compiled functions instead.


titzer (commented):
Would it be possible to make these into `.wast` tests? That will allow running them on the reference interpreter and verifying their outputs. That will require some scripting to do that for all these tests. 

Also, they shouldn't fail *yet*, so maybe we don't need the failures file yet?

matthew-mojira:
It fails because the `--fast-functions` option it uses also isn't added yet. I could probably add it as a dormant flag in another PR.

matthew-mojira:
> Would it be possible to make these into `.wast` tests? That will allow running them on the reference interpreter and verifying their outputs. That will require some scripting to do that for all these tests.

Some tests use `wave.new_func`. How should that be incorporated in the `.wast` version?

titzer:
> > Would it be possible to make these into `.wast` tests? That will allow running them on the reference interpreter and verifying their outputs. That will require some scripting to do that for all these tests.
> 
> Some tests use `wave.new_func`. How should that be incorporated in the `.wast` version?

Ah, good question. Let's start by separating out everything that doesn't need `wave.new_func` into `.wast` tests and leave the rest as standalone. I.e. try to test the functionality of fast handlers as best as possible independent of JIT.

After the JIT interface proposal has an implementation in the reference interpreter, then those tests can be converted to use `func.new`--that'll be good because they can double as spec tests.

titzer:
> It fails because the `--fast-functions` option it uses also isn't added yet. I could probably add it as a dormant flag in another PR.

Go ahead and add the flag, even though it does nothing.

titzer (approved):

PR #653, opened 2026-06-17
Move restoreDispatchTableReg and computePc to MacroAssembler

matthew-mojira:


src/engine/x86-64/X86_64Interpreter.v3
```diff
@@ -561,7 +561,7 @@ class X86_64InterpreterGen(ic: X86_64InterpreterCode, w: DataWriter) {
 			masm.emit_v3_X86_64Stack_rsp_r_r(xenv.sp, xenv.scratch);
 			// Restore interpreter registers
 			restoreCallerIVars();
-			restoreDispatchTableReg();
+			if (!FeatureDisable.globalProbes) masm.emit_restore_dispatch_table_reg(xenv.dispatch);
```
titzer: I think this is better if we keep this routine (and it's factored-out flag check) and just have it call the macro assembler method.

titzer (approved):

PR #678, opened 2026-09-01
Add `FAST_CALL` instructions

matthew-mojira:
The fast call instructions are internal and not handled by the code validator, so any Wasm binary that tries to use it will not validate (and that case should never be supported since the validator would rewrite `CALL` into `FAST_CALL`)


src/engine/Opcodes.v3
```diff
@@ -1129,6 +1173,100 @@ component Opcodes {
 			}
 		}
 	}
+	def indexToFastCall(index: int) -> Opcode {
```
titzer, line 1178: I think you could move this into `Opcodes` and do this with a bounds check and an array access into the internal opcodes array already built there.

src/engine/Opcodes.v3
```diff
@@ -1129,6 +1173,100 @@ component Opcodes {
...
+			35 => op = Opcode.FAST_CALL35;
+			36 => op = Opcode.FAST_CALL36;
+			37 => op = Opcode.FAST_CALL37;
+			38 => op = Opcode.FAST_CALL38;
+			39 => op = Opcode.FAST_CALL39;
+			_ => System.error("indexToFastCall", "out of range");
+		}
+		return op;
+	}
+	def fastCallToIndex(op: Opcode) -> int {
```
titzer, line 1182: Similarly, this could move to `Opcodes` and do a bounds check and a subtraction.

PR #691, opened 2026-09-15
Add state tracking for interpreter registers used for fast handlers

matthew-mojira:


src/engine/compiler/SinglePassCompiler.v3
```diff
@@ -60,6 +60,7 @@ component SpcConsts {
 	def kindToFlags(kind: ValueKind) -> byte {
 		return byte.view(kind.tag) << 4;
 	}
+	def IVAR_SLOT = TMP_SLOT + 1;
```
titzer: Move this up to be with the other constant fields.

src/engine/compiler/SinglePassCompiler.v3
```diff
@@ -2541,6 +2547,7 @@ class SinglePassCompiler(xenv: SpcExecEnv, masm: MacroAssembler, regAlloc: RegAl
 		return regAlloc.alloc(kind, slot);
 	}
 	def addSpillCost(score: int, reg: Reg, slot: int) -> int {
+		if (slot == IVAR_SLOT) return 50 + score;             // fast interpreter register?
```
titzer, line 2550: We could also differentiate the cost for ones that need to be saved versus just reloaded.

src/engine/compiler/SinglePassCompiler.v3
```diff
@@ -2667,6 +2674,7 @@ class SinglePassCompiler(xenv: SpcExecEnv, masm: MacroAssembler, regAlloc: RegAl
 	}
 	def emitSpill(slot: int) {
 		if (slot == TMP_SLOT) return;
+		if (slot == IVAR_SLOT) return;
```
titzer, line 2678: For the dirty ivars (e.g. IP), we need to spill them to their slot.

src/engine/compiler/SinglePassCompiler.v3
```diff
@@ -3216,6 +3228,19 @@ class SpcState(regAlloc: RegAlloc) {
...
 	}
+	def clear() {
+		if (useIVars) regAlloc.forEachAllocatedReg(NO_REG, freeExceptIVar);
+		else regAlloc.clear();
+	}
+	def freeExceptIVar(reg: Reg) {
+		if (!holdsIVar(reg)) regAlloc.free(reg);
+	}
+	def holdsIVar(reg: Reg) -> bool {
+		if (!regAlloc.unassign(reg, IVAR_SLOT)) return false;
```
titzer: This is probably worth a new query method on the register allocator to avoid unassigning and reassigning.

src/engine/x86-64/X86_64MasmRegs.v3
```diff
@@ -185,6 +190,14 @@ component X86_64MasmRegs {
 		];
 		return RegAlloc.new(CONFIG.poolMap, pools, null);
 	})();
+	// A register allocator for interpreter contexts for fast handlers.
+	def FAST_INT_ALLOC = (fun -> RegAlloc {
```
titzer, line 197: If possible we should build the list of registers by using the previously-defined int exec environment, etc; otherwise this will get out of date if we shuffle assignments around.

titzer (approved):

PR #700, opened 2026-09-18
Dispatch on the local declarations before emitting entry probe in SPC

matthew-mojira:
When a function calls another function and it gets inlined, a whamm probe attached to the callee function entry (that is also being inlined) corrupts the state of the `BytecodeIterator` that the SPC uses, causing it to misread the locals section for the call to `dispatchLocalDecls` (as having no locals). This PR moves the dispatch before the entry probe so the `BytecodeIterator` can properly see the local declarations.


titzer (approved):
Nice catch!

